PR #5, opened 2019-06-06
(#1713159) rules: skip memory hotplug on ppc64

jsynacek:
Resolves (#1713159)


haraldh (commented):
Wouldn't that better be in an architecture dependent separate rules file, to be ifdef'ed in the spec file?

msekletar (approved):
LGTM

msekletar:
And I agree with Harald that we should start splitting 40-redhat.rules (we tried in the past and it didn't quite work out though).

Memory on-lining policy should be a separate rule file and we would include it only on architectures where we want on-lining to happen automatically.

PR #6, opened 2019-06-11
(#1719153) (#1719153) cryptsetup token support backport

jsynacek:


msekletar (approved):
LGTM (but quite frankly I don't really understand all the details however I trust Milan).

I'd feel better about this if we added test case for token support with LUKS2.

PR #7, opened 2019-06-12
(#1719577) core: implement per unit journal rate limiting

dtardon:
Add LogRateLimitIntervalSec= and LogRateLimitBurst= options for
services. If provided, these values get passed to the journald
client context, and those values are used in the rate limiting
function in the journal over the the journald.conf values.

Part of #10230

(cherry picked from commit 90fc172e191f44979005a524521112f2bd1ff21b)

Resolves: #1719577


msekletar (approved):
LGTM

@dtardon thank you very much for backporting this. It is much appreciated!

PR #17, opened 2019-09-03
(#1748258) buildsys: don't garbage collect sections while linking

msekletar:
gc-sections is actually very aggressive and garbage collects ELF
sections used by annobin gcc plugin and annocheck then reports gaps in
coverage. Let's drop that linker flag.

RHEL-only

Resolves: #1748258


jsynacek (approved):
LGTM. 100kB increase in size is negligible.

PR #26, opened 2019-10-11
(#1753369) udev: enable U2F

dtardon:


rules/60-fido-id.rules
```diff
@@ -0,0 +1,7 @@
+# do not edit this file, it will be overwritten on update
+
+ACTION=="remove", GOTO="fido_id_end"
```
jsynacek, line 3: Shouldn't this also say `SUBSYSTEM="hidraw"` to not match on every single device and needlessly execute a goto that does nothing?

dtardon: I checked a handful of randomly picked rules files and they are all start with the same rule (modulo the goto name). So presumably it's not a problem?

jsynacek: Well, they're all inefficient in the same way, in my opinion. But the impact is probably negligible, so let's leave it as it is.

jsynacek (approved):
LGTM.

PR #28, opened 2019-10-17
(#1762679) udev: introduce CONST key name

jsynacek:
Currently, there is no way to match against system-wide constants, such
as architecture or virtualization type, without forking helper binaries.
That potentially results in a huge number of spawned processes which
output always the same answer.

This patch introduces a special CONST keyword which takes a hard-coded
string as its key and returns a value assigned to that key. Currently
implemented are CONST{arch} and CONST{virt}, which can be used to match
against the system's architecture and virtualization type.

(based on commit 4801d8afe2ff1c1c075c9f0bc5631612172e0bb7)

Resolves: #1762679

This patch also includes https://github.com/systemd/systemd/pull/13790, which I forgot to add to the original upstream PR.

### DIFF
--- a/man/udev.xml
+++ b/man/udev.xml
@@ -237,6 +237,32 @@
         </varlistentry>
 
         <varlistentry>
+          <term><varname>CONST{<replaceable>key</replaceable>}</varname></term>
+          <listitem>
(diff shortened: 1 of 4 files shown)


dtardon (approved):
LGTM

PR #29, opened 2019-10-20
(#1763161) path: stop watching path specs once we triggered the target unit

msekletar:
We start watching them again once we get a notification that triggered
unit entered inactive or failed state.

Fixes: #10503
(cherry picked from commit 8fca6944c2ee20c63d62154c8badddc77170b176)

Resolves: #1763161


jsynacek (approved):
LGTM.

PR #32, opened 2019-10-21
(#1763619) journald: fixed assertion failure when system journal rotation fails …

dtardon:
…(#9893)

(cherry picked from commit fd790d6f09b10a87b007b71403cb018f18ff91c9)

Resolves: #1763619


jsynacek (approved):
LGTM.

PR #33, opened 2019-10-21
(#1763602) core: reduce the number of stalled PIDs from the watched processes li…

dtardon:
…st when possible

Some PIDs can remain in the watched list even though their processes have
exited since a long time. It can easily happen if the main process of a forking
service manages to spawn a child before the control process exits for example.

However when a pid is about to be mapped to a unit by calling unit_watch_pid(),
the caller usually knows if the pid should belong to this unit exclusively: if
we just forked() off a child, then we can be sure that its PID is otherwise
unused. In this case we take this opportunity to remove any stalled PIDs from
the watched process list.

If we learnt about a PID in any other form (for example via PID file, via
searching, MAINPID= and so on), then we can't assume anything.

(cherry picked from commit f75f613d259e9332be8e9657fa37e73f7aabcb8a)

Resolves: #1763602

### DIFF
--- a/src/core/cgroup.c
+++ b/src/core/cgroup.c
@@ -1929,7 +1929,7 @@ static int unit_watch_pids_in_path(Unit *u, const char *path) {
                 pid_t pid;
 
                 while ((r = cg_read_pid(f, &pid)) > 0) {
-                        r = unit_watch_pid(u, pid);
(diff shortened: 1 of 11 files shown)


jsynacek (approved):
LGTM.

dtardon:
This commit has already been merged via another PR...

PR #39, opened 2019-11-09
(#1829798) core: coldplug possible nop_job

chenglin130:
When a unit in a state INACTIVE or DEACTIVATING, JobType JOB_TRY_RESTART or
JOB_TRY_RELOAD will be collapsed to JOB_NOP. And use u->nop_job instead
of u->job.

If a JOB_NOP job is going on with a waiting state, a parallel daemon-reload
just install it during deserialization. Without a coldplug, the job will
not be in m->run_queue, which results in a hung try-restart or
try-reload process.

Reproduce:
1. run systemctl try-restart test.servcie (inactive) repeatly in a terminal.
2. run systemctl daemon-reload repeatly in other terminals.

After successful reproduce, systemctl list-jobs will list the hang job.

Upsteam:
systemd/systemd#13124

### DIFF
--- a/src/core/unit.c
+++ b/src/core/unit.c
@@ -3755,6 +3755,7 @@ int unit_add_node_dependency(Unit *u, const char *what, bool wants, UnitDependen
 int unit_coldplug(Unit *u) {
         int r = 0, q;
         char **i;
+        Job *uj;
 
         assert(u);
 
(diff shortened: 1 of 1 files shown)


jsynacek:
@chenglin130 Is there a bugzilla that will be fixed by this PR? Every PR here has to have a corresponding bugzilla filed for the correct RHEL version.

chenglin130:
@jsynacek 
I found this bug on centos7.6, and have tested the solution on centos 7.6. It's not difficult to reproduce this bug. By the code, I have confirmed that this bug also exists on rhel8.

And I found the Upstream fix this bug already. So I think RHEL also need this patch to fix it. I don't know if there is a record for it on bugzilla.

jsynacek:
Merged in https://github.com/systemd-rhel/rhel-8/commit/4a1405998671caaaad5b24d4cef309c05175b1c1.

PR #40, opened 2019-11-11
(#1687512) (#1687512) core: imply NNP and SUID/SGID restriction for DynamicUser=yes service

jsynacek:
Let's be safe, rather than sorry. This way DynamicUser=yes services can
neither take benefit of, nor create SUID/SGID binaries.

Given that DynamicUser= is a recent addition only we should be able to
get away with turning this on, even though this is strictly speaking a
binary compatibility breakage.

(cherry picked from commit bf65b7e0c9fc215897b676ab9a7c9d1c688143ba)
Resolves: #1687512


jsynacek:
v2.

src/shared/seccomp-util.c
```diff
@@ -1747,3 +1749,174 @@ int seccomp_lock_personality(unsigned long personality) {
 
         return 0;
 }
+
+int seccomp_protect_hostname(void) {
```
msekletar: This function was probably backported by accident. 

I am fine with including it but we also need to backport corresponding parser, man page blurp, etc ...

msekletar (changes requested):
LGTM, except the ProtectHostname= problem.

src/shared/seccomp-util.c
jsynacek: Right, this was by accident. I'm going to remove it.

jsynacek:
v3.

msekletar (approved):
LGTM

PR #41, opened 2019-11-18
(#1734787) core: introduce NUMAPolicy and NUMAMask options

msekletar:
This PR also contains a bunch of preparatory work on CPUSet APIs.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -88,46 +88,52 @@ static enum {
         ACTION_DUMP_CONFIGURATION_ITEMS,
         ACTION_DUMP_BUS_PROPERTIES,
 } arg_action = ACTION_RUN;
-static char *arg_default_unit = NULL;
-static bool arg_system = false;
-static bool arg_dump_core = true;
-static int arg_crash_chvt = -1;
-static bool arg_crash_shell = false;
-static bool arg_crash_reboot = false;
-static char *arg_confirm_spawn = NULL;
-static ShowStatus arg_show_status = _SHOW_STATUS_UNSET;
-static bool arg_switched_root = false;
-static bool arg_no_pager = false;
-static bool arg_service_watchdogs = true;
+
+/* Those variables are initalized to 0 automatically, so we avoid uninitialized memory access.
+ * Real defaults are assigned in reset_arguments() below. */
+static char *arg_default_unit;
+static bool arg_system;
+static bool arg_dump_core;
+static int arg_crash_chvt;
+static bool arg_crash_shell;
+static bool arg_crash_reboot;
+static char *arg_confirm_spawn;
+static ShowStatus arg_show_status;
+static bool arg_switched_root;
+static bool arg_no_pager;
+static bool arg_service_watchdogs;
 static char ***arg_join_controllers = NULL;
-static ExecOutput arg_default_std_output = EXEC_OUTPUT_JOURNAL;
-static ExecOutput arg_default_std_error = EXEC_OUTPUT_INHERIT;
-static usec_t arg_default_restart_usec = DEFAULT_RESTART_USEC;
-static usec_t arg_default_timeout_start_usec = DEFAULT_TIMEOUT_USEC;
-static usec_t arg_default_timeout_stop_usec = DEFAULT_TIMEOUT_USEC;
-static usec_t arg_default_start_limit_interval = DEFAULT_START_LIMIT_INTERVAL;
-static unsigned arg_default_start_limit_burst = DEFAULT_START_LIMIT_BURST;
-static usec_t arg_runtime_watchdog = 0;
-static usec_t arg_shutdown_watchdog = 10 * USEC_PER_MINUTE;
-static char *arg_watchdog_device = NULL;
-static char **arg_default_environment = NULL;
-static struct rlimit *arg_default_rlimit[_RLIMIT_MAX] = {};
-static uint64_t arg_capability_bounding_set = CAP_ALL;
-static bool arg_no_new_privs = false;
-static nsec_t arg_timer_slack_nsec = NSEC_INFINITY;
-static usec_t arg_default_timer_accuracy_usec = 1 * USEC_PER_MINUTE;
-static Set* arg_syscall_archs = NULL;
-static FILE* arg_serialization = NULL;
-static bool arg_default_cpu_accounting = false;
-static bool arg_default_io_accounting = false;
-static bool arg_default_ip_accounting = false;
-static bool arg_default_blockio_accounting = false;
-static bool arg_default_memory_accounting = MEMORY_ACCOUNTING_DEFAULT;
-static bool arg_default_tasks_accounting = true;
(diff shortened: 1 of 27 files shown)


msekletar:
CI failure (ninja test) seems unrelated,

```
Assertion 'sched_setscheduler(0, SCHED_FIFO, &(struct sched_param) { .sched_priority = 1 }) >= 0' failed at ../src/test/test-seccomp.c:354, function test_restrict_realtime(). Aborting.
realtimeseccomp terminated by signal ABRT.
Assertion 'wait_for_terminate_and_check("realtimeseccomp", pid, WAIT_LOG) == EXIT_SUCCESS' failed at ../src/test/test-seccomp.c:374, function test_restrict_realtime(). Aborting.
```

mrc0mmand:
> CI failure (ninja test) seems unrelated,
> 
> ```
> Assertion 'sched_setscheduler(0, SCHED_FIFO, &(struct sched_param) { .sched_priority = 1 }) >= 0' failed at ../src/test/test-seccomp.c:354, function test_restrict_realtime(). Aborting.
> realtimeseccomp terminated by signal ABRT.
> Assertion 'wait_for_terminate_and_check("realtimeseccomp", pid, WAIT_LOG) == EXIT_SUCCESS' failed at ../src/test/test-seccomp.c:374, function test_restrict_realtime(). Aborting.
> ```

It's interesting though that it happened again after rescheduling the job :thinking:

dtardon (approved):
LGTM

mrc0mmand:
> > CI failure (ninja test) seems unrelated,
> > ```
> > Assertion 'sched_setscheduler(0, SCHED_FIFO, &(struct sched_param) { .sched_priority = 1 }) >= 0' failed at ../src/test/test-seccomp.c:354, function test_restrict_realtime(). Aborting.
> > realtimeseccomp terminated by signal ABRT.
> > Assertion 'wait_for_terminate_and_check("realtimeseccomp", pid, WAIT_LOG) == EXIT_SUCCESS' failed at ../src/test/test-seccomp.c:374, function test_restrict_realtime(). Aborting.
> > ```
> 
> It's interesting though that it happened again after rescheduling the job thinking

I triggered a CI job on `master` and the issue doesn't occur there.

src/test/test-cpu-set-util.c
```diff
@@ -181,6 +181,36 @@ static void test_parse_cpu_set_extend(void) {
...
+        log_info("/* %s */", __func__);
+
+        assert_se(parse_cpu_set_extend("1 3 8 100-200", &c, true, NULL, "fake", 1, "CPUAffinity") == 0);
+        assert_se(s = cpu_set_to_string(&c));
+        log_info("cpu_set_to_string: %s", s);
+        assert_se(CPU_COUNT_S(c.allocated, c.set) == 104);
+
+        _cleanup_free_ uint8_t *array = NULL;
+        size_t allocated;
+        static const char expected[32] =
```
jsynacek, line 250: Isn't this going to fail with different-than-expected endianness?

jsynacek (commented):
See inline comment.

jsynacek (approved):
Anyway, looks good to me. If the tests are wrong, we will know.

msekletar:
I can't reproduce the CI failure. I ran the `ninja test` manually multiple times. I've also tried to run it as systemd service via `systemd-run` and it always passed.

I am just tempted to waive the results and move on...Although, I'd prefer if somebody else did it (after double-checking) since this is my PR.

@jsynacek @dtardon can you have a look one more time and waive (ci-waive label) if appropriate?

jsynacek:
I consider the seccomp failures in `test_restrict_realtime()` irrelevant to this pull request. In my opinion, you can waive the CI failures.

@mrc0mmand Do we have access to that machine? I wonder if the correct realtime kernel options are even enabled.

msekletar:
@jsynacek Yesterday we figured out together with @mrc0mmand what is going on. Basically, in CI `ninja test` was running in a session scope unit that didn't have any realtime budget assigned and syscall was failing with EPERM.

I was not able to reproduce because in my case (minimal CentOS 8) I didn't have CPUAccounting enabled  and hence test suite ran in `/` cgroup of cpu controller and that has all budget by default and therefore syscall was successful.

src/core/main.c
```diff
@@ -1935,30 +1983,74 @@ static int do_queue_default_job(
...
+        arg_capability_bounding_set = CAP_ALL;
+        arg_no_new_privs = false;
+        arg_timer_slack_nsec = NSEC_INFINITY;
+        arg_default_timer_accuracy_usec = 1 * USEC_PER_MINUTE;
+
         arg_syscall_archs = set_free(arg_syscall_archs);
+
+        /* arg_serialization — ignore */
+
+        arg_default_cpu_accounting = -1;
```
mrc0mmand: This line right here is the culprit of the `test-seccomp` failure. Even though `arg_default_cpu_accounting` is zeroed-out, as it's in the BSS section, after the first systemd reload it's set to `-1` it, which equals to `true` after the boolean conversion, thus the CPU accounting is enabled system-wide.

The main issue here is that the CPU accounting is "forcefully" enabled even if you don't change the default in `/etc/systemd/system.conf`, because the respective line is commented out:

```
# grep DefaultCPUAccounting /etc/systemd/system.conf
#DefaultCPUAccounting=no
```
thus the unexpected behavior.

This can be easily reproduced, for example, in our `1minutetip`:

```
$ 1minutetip --flavor ci.c1.xlarge -n rhel8
...
## Following commands are executed inside the VM
# git clone https://github.com/systemd-rhel/rhel-8
# cd rhel-8/
# git fetch -fu origin refs/pull/41/merge:pr
# meson build
# ninja -C build install
# dracut -f --regenerate-all
# systemctl reboot
...
# systemctl show -p DefaultCPUAccounting
DefaultCPUAccounting=yes
# grep DefaultCPUAccounting /etc/systemd/system.conf
#DefaultCPUAccounting=no
```

mrc0mmand (changes requested):

src/core/main.c
mrc0mmand: I suspect this may have been fixed in https://github.com/systemd/systemd/pull/10507/files#diff-108b33cf1bd0765d116dd401376ca356R755 where in case of `arg_default_cpu_accounting == -1` it would get overridden by a value returned from `cpu_accounting_is_cheap()`, which should return false when running on a system with cgroups v1.

mrc0mmand (approved):

mrc0mmand:
As discussed with @msekletar, the missing tri-state logic for `arg_default_cpu_accounting` is not something we really need, and even want, in RHEL 8, so the last patch, which simply brings back the original behavior is more than enough to fix the regression (and the `test-seccomp` test).

Also, the tri-state logic is useful only when the system runs on full cgroups v2 (see [the explanation](https://github.com/systemd/systemd/pull/10507/files#diff-0d48d4d5663167a0bb6786a4abaddc35R2830-R2856)), so we don't really miss anything, at least in RHEL 8.

PR #42, opened 2019-11-27
(#1777037) ask-password: prevent buffer overrow when reading from keyring

dtardon:
When we read from keyring, a temporary buffer is allocated in order to
determine the size needed for the entire data. However, when zeroing that area,
we use the data size returned by the read instead of the lesser size allocate
for the buffer.

That will cause memory corruption that causes systemd-cryptsetup to crash
either when a single large password is used or when multiple passwords have
already been pushed to the keyring.

Signed-off-by: Thadeu Lima de Souza Cascardo <cascardo@canonical.com>
(cherry picked from commit 59c55e73eaee345e1ee67c23eace8895ed499693)
(cherry picked from commit c6c8e0d097d6ba12471c6112c3fd339ea40329d5)

Resolves: #1777037

### DIFF
--- a/src/shared/ask-password-api.c
+++ b/src/shared/ask-password-api.c
@@ -79,7 +79,7 @@ static int retrieve_key(key_serial_t serial, char ***ret) {
                 if (n < m)
                         break;
 
-                explicit_bzero(p, n);
+                explicit_bzero(p, m);
                 free(p);
                 m *= 2;
         }


msekletar:
Hmm, not quite sure what is going on with CI. Failures are quite serious in my opinion (ninja test failed and boot timed out for CRYPSETUP test case), though, they seem completely unrelated to this PR.

msekletar (approved):
LGTM

msekletar:
I've waived the CI results after talking to @mrc0mmand as he confirmed my suspicion that failures are unrelated.

PR #45, opened 2019-12-02
(#1696602) cryptsetup: reduce the chance that we will be OOM killed

msekletar:
cryptsetup introduced optional locking scheme that should serialize
unlocking keyslots which use memory hard key derivation
function (argon2). Using the serialization should prevent OOM situation
in early boot while unlocking encrypted volumes.

(cherry picked from commit 408c81f62454684dfbff1c95ce3210d06f256e58)

Resolves: #1696602


jsynacek (approved):
LGTM.

PR #56, opened 2020-01-07
(#1788085) fix false journald assert

jsynacek:


mrc0mmand:
Ignore the CentOS CI fail for now, it's already under investigation, will re-trigger it after the issue is resolved.

dtardon (approved):
LGTM

jsynacek:
The CentOS CI failure looks like some kind of internal error and totally unrelated to the patch.

mrc0mmand:
> The CentOS CI failure looks like some kind of internal error and totally unrelated to the patch.

Whoops, sorry, my bad. I actually fixed the CI issue but forgot to reschedule the job. Done it now and it seems to be running well, results should be up in ~10 minutes.

PR #66, opened 2020-02-19
(#1798504) bus_open leak sd_event_source when udevadm trigger。

dtardon:
On my host, when executing the udevadm trigger, I only receive the change event, which causes memleak

(cherry picked from commit b2774a3ae692113e1f47a336a6c09bac9cfb49ad)

Resolves: #1798504


jsynacek (approved):
LGTM.

PR #67, opened 2020-02-20
(#1804252) meson: allow setting the version string during configuration

jsynacek:
Very loosely based on upstream commits e1ca734edd17a90a325d5b566a4ea96e66c206e5
and 681bd2c524ed71ac04045c90884ba8d55eee7b66.

Resolves: #1804252

### DIFF
--- a/src/udev/udev-ctrl.c
+++ b/src/udev/udev-ctrl.c
@@ -239,7 +239,10 @@ static int ctrl_send(struct udev_ctrl *uctrl, enum udev_ctrl_msg_type type, int 
         int err = 0;
 
         memzero(&ctrl_msg_wire, sizeof(struct udev_ctrl_msg_wire));
-        strcpy(ctrl_msg_wire.version, "udev-" PACKAGE_VERSION);
+        /* jsynacek: As PACKAGE_VERSION is populated from the spec file with %{version}-%{release},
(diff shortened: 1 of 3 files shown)


src/udev/udev-ctrl.c
```diff
@@ -239,7 +239,10 @@ static int ctrl_send(struct udev_ctrl *uctrl, enum udev_ctrl_msg_type type, int
         int err = 0;
 
         memzero(&ctrl_msg_wire, sizeof(struct udev_ctrl_msg_wire));
-        strcpy(ctrl_msg_wire.version, "udev-" PACKAGE_VERSION);
+        /* jsynacek: As PACKAGE_VERSION is populated from the spec file with %{version}-%{release},
+         * it might not fit entirely into the version field. */
+        strncpy(ctrl_msg_wire.version, "udev-" PACKAGE_VERSION, 16-5);
```
jsynacek, line 244: There were two possibilites how to solve this:
1) Use `udev-<project version>`, in our case "udev-239", as upstream does.
2) Use the new versioning scheme limited by the size of the .version field.

I decided to go with 2), as it also makes use of the new functionality. I'm not sure how it matters in this case.

dtardon (approved):
LGTM

PR #73, opened 2020-03-09
(#1808940) cgroup: make sure that cpuset is supported on cgroup v2 and disabled with v1

msekletar:
Resolves: #1808940


jsynacek (approved):
LGTM.

PR #75, opened 2020-03-18
(#1797679) logind: check PolicyKit before allowing VT switch

dtardon:
Let's lock this down a bit. Effectively nothing much changes, since the
default PK policy will allow users on the VT to change VT. Only users
with no local VT session won't be able to switch VTs.

(cherry picked from commit 4acf0cfd2f92edb94ad48d04f1ce6c9ab4e19d55)

Resolves: #1797679


jsynacek (approved):
LGTM.

PR #80, opened 2020-04-06
(#1816908) man: be clearer that .timer time expressions need to be reset to over…

dtardon:
…ride them

let's be clearer about the overriding concept for OnCalendar= settings.

Prompted by this thread:

https://lists.freedesktop.org/archives/systemd-devel/2019-March/042351.html
(cherry picked from commit 58031d99c6320855b86f4890baa9165597e3d841)

Resolves: #1816908


jsynacek (approved):
LGTM.

PR #85, opened 2020-04-14
(#1823767) Testsuite backports

mrc0mmand:
`TEST-24-UNIT-TESTS` is completely broken and what's worse - the test fail was masked as well.

The first patch resolves the return code masking issue, whereas the rest fixes the uncovered test errors. Also, the patchset extends the test coverage, so now the unit tests are run under QEMU and in both privileged and unprivileged systemd-nspawn containers. Lastly, I pulled down my parallelization patch from upstream which significantly speeds up the test on multi-core machines (in CentOS CI the runtime went from 21 minutes to 14 minutes).

The last patch is not relevant to `TEST-24-UNIT-TEST` by itself, but it's going to help with https://github.com/systemd/systemd-centos-ci/pull/242.

Resolves: #1823767

### DIFF
--- a/catalog/meson.build
+++ b/catalog/meson.build
@@ -17,7 +17,6 @@ in_files = '''
 
 support_url = get_option('support-url')
 support_sed = 's~%SUPPORT_URL%~@0@~'.format(support_url)
-build_catalog_dir = meson.current_build_dir()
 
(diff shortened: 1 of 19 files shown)


jsynacek (approved):
LGTM.

PR #92, opened 2020-05-15
(#1835594) resolvconf: fixes for the compatibility interface

dtardon:
Also use compat_main() when called as `resolvconf`, since the interface
is closer to that of `systemd-resolve`.

Use a heap allocated string to set arg_ifname, since a stack allocated
one would be lost after the function returns. (This last one broke the
case where an interface name was suffixed with a dot, such as in
`resolvconf -a tap0.dhcp`.)

Tested:
  $ build/resolvconf -a nonexistent.abc </etc/resolv.conf
  Unknown interface 'nonexistent': No such device

Fixes #9423.

(cherry picked from commit 5a01b3f35d7b6182c78b6973db8d99bdabd4f9c3)

Resolves: #1835594

### DIFF
--- a/src/resolve/resolvconf-compat.c
+++ b/src/resolve/resolvconf-compat.c
@@ -53,6 +53,8 @@ static int parse_nameserver(const char *string) {
 
                 if (strv_push(&arg_set_dns, word) < 0)
                         return log_oom();
+
+                word = NULL;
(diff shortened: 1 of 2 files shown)


jsynacek (approved):
LGTM.

PR #93, opened 2020-05-15
(#1836024) tmpfiles: fix crash with NULL in arg_root and other fixes and tests

dtardon:
The function to replacement paths into the configuration file list was borked.
Apart from the crash with empty root prefix, it would incorrectly handle the
case where root *was* set, and the replacement file was supposed to override
an existing file.

prefix_root is used instead of path_join because prefix_root removes duplicate
slashes (when --root=dir/ is used).

A test is added.

Fixes #11124.

(cherry picked from commit 082bb1c59bd4300bcdc08488c94109680cfadf57)

Resolves: #1836024

### DIFF
--- a/src/basic/conf-files.c
+++ b/src/basic/conf-files.c
@@ -204,14 +204,17 @@ int conf_files_insert(char ***strv, const char *root, char **dirs, const char *p
                 if (c == 0) {
                         char **dir;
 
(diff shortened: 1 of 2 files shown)


jsynacek (approved):
LGTM.

PR #98, opened 2020-05-19
(#1837094) nspawn: move payload to sub-cgroup first, then sync cgroup trees

mrc0mmand:
if we sync the legacy and unified trees before moving to the right
subcgroup then ultimately the cgroup paths in the hierarchies will be
out-of-sync... Hence, let's move the payload first, and sync then.

Addresses: https://github.com/systemd/systemd/pull/9762#issuecomment-441187979
(cherry picked from commit 27da7ef0d09e00eae821f3ef26e1a666fe7aa087)

Resolves: #1837094

---

Setting `dont-merge` label, as I'd like to give it a try in https://github.com/systemd/systemd-centos-ci/pull/257


mrc0mmand:
Given the results in systemd/systemd-centos-ci#257 it indeed resolves the issue described in BZ#1837094 on both legacy & unified hierarchies.

dtardon (approved):
LGTM

PR #100, opened 2020-05-19
(#1837423) nspawn: move payload to sub-cgroup first, then sync cgroup trees

mrc0mmand:
if we sync the legacy and unified trees before moving to the right
subcgroup then ultimately the cgroup paths in the hierarchies will be
out-of-sync... Hence, let's move the payload first, and sync then.

Addresses: https://github.com/systemd/systemd/pull/9762#issuecomment-441187979
(cherry picked from commit 27da7ef0d09e00eae821f3ef26e1a666fe7aa087)
(cherry picked from commit 8ee1465520ad49892a0a378626ef93abc03f4d4e)

Resolves: #1837423


jsynacek (approved):
LGTM.

PR #104, opened 2020-05-26
(#1748840) mount: don't add Requires for tmp.mount

dtardon:
This is a follow-up to #1619292.

rhel-only
Resolves: #1748840


jsynacek (approved):
LGTM.

PR #108, opened 2020-06-15
(#1737283) core: don't consider SERVICE_SKIP_CONDITION for abnormal or failure r…

dtardon:
…estarts

Fixes: #16115
(cherry picked from commit bb9244781c6fc7608f7cac910269f8987b8adc01)

Related: #1737283

PR #109, opened 2020-06-15
(#1830861) Freezer backport

msekletar:


msekletar:
@mrc0mmand I haven't backported the integration test because upstream infrastructure to run integration tests was substantially reworked and tests already relies on this rework.

I can either amend it so that it is run the "old-fashioned" way or skip backporting it altogether (assuming we will use most of the test code in the standalone RHEL integration test we would create). I have a slight preference towards the second option.

mrc0mmand:
> @mrc0mmand I haven't backported the integration test because upstream infrastructure to run integration tests was substantially reworked and tests already relies on this rework.
> 
> I can either amend it so that it is run the "old-fashioned" way or skip backporting it altogether (assuming we will use most of the test code in the standalone RHEL integration test we would create). I have a slight preference towards the second option.

I'd prefer backporting the test and adapting it to the *pre-rewrite* state as for example @dtardon did in #108. I'd rather have a test in CentOS CI covering this feature to some degree where I can be fairly certain it DTRT than relying completely on the mess we call infrastructure internally...

mrc0mmand:
Also:

```
[[0;32m  OK  [0m] Started Create list of required sta…vice nodes for the current kernel.
[0;1;31mAssertion 'u' failed at ../src/core/cgroup.c:2240, function unit_add_to_cgroup_empty_queue(). Aborting.[0m
[0;1;31mCaught <ABRT>, dumped core as pid 315.[0m
[0;1;31mFreezing execution.[0m
```

Full log: https://ci.centos.org/job/systemd-rhel8-pr-build/448/artifact//systemd-centos-ci/artifacts_all/artifacts_1dIJE5/bootstrap-logs-rhel8.BxW/sanity-boot-check.log

msekletar:
Ok, I will backport the test. As for the crash, `WIP:` in summary of the PR is there for a reason 😬

msekletar:
@dtardon @jsynacek this is ready for review. CentOS CI still fails because there is a still problem with my rework of the integration test. However, when I run test manually it passes.

test/TEST-38-FREEZER/test.sh
```diff
@@ -0,0 +1,50 @@
...
+test_setup() {
+    create_empty_image
+    mkdir -p $TESTDIR/root
+    mount ${LOOPDEV}p1 $TESTDIR/root
+
+    (
+        LOG_LEVEL=5
+        eval $(udevadm info --export --query=env --name=${LOOPDEV}p2)
+
+        setup_basic_environment
```
mrc0mmand, line 16: I suspect calling

```
dracut_install tr cut
```

should do the trick to fix:


```
Jun 16 14:54:36 systemd-testsuite testsuite.sh[595]: ++ busctl get-property org.freedesktop.systemd1 /org/freedesktop/systemd1/unit/testsuite_2d38_2dsleep_2eservice org.freedesktop.systemd1.Unit FreezerState
Jun 16 14:54:36 systemd-testsuite testsuite.sh[595]: ++ cut -d ' ' -f2
Jun 16 14:54:36 systemd-testsuite testsuite.sh[595]: ++ tr -d '"'
Jun 16 14:54:36 systemd-testsuite testsuite.sh[595]: /testsuite.sh: line 89: tr: command not found
Jun 16 14:54:36 systemd-testsuite testsuite.sh[595]: /testsuite.sh: line 89: cut: command not found
```

msekletar:
@mrc0mmand good catch. Thank you!

test/TEST-38-FREEZER/test.sh
```diff
@@ -0,0 +1,51 @@
...
+        draut_install tr cut
```
mrc0mmand:
```suggestion
        dracut_install tr cut
```

msekletar: Ahh 🙈

msekletar:
@mrc0mmand do you know what is going on with Travis CI? It seems like Xenial image got updated and no longer contains `diff` binary...weird.

mrc0mmand:
@msekletar *sigh*, looks like a few hours ago a new CentOS 8 Docker image went out and they apparently dropped the `diffutils` package from it. If you could sneak in this patch, it would be great:

```diff
$ git diff
diff --git a/ci/travis-centos-rhel8.sh b/ci/travis-centos-rhel8.sh
index a1502e15ee..cd0857fd29 100755
--- a/ci/travis-centos-rhel8.sh
+++ b/ci/travis-centos-rhel8.sh
@@ -15,7 +15,7 @@ CONT_NAME="${CONT_NAME:-centos-$CENTOS_RELEASE-$RANDOM}"
 DOCKER_EXEC="${DOCKER_EXEC:-docker exec -it $CONT_NAME}"
 DOCKER_RUN="${DOCKER_RUN:-docker run}"
 REPO_ROOT="${REPO_ROOT:-$PWD}"
-ADDITIONAL_DEPS=(libasan libubsan net-tools strace nc e2fsprogs quota dnsmasq)
+ADDITIONAL_DEPS=(libasan libubsan net-tools strace nc e2fsprogs quota dnsmasq diffutils)
 # RHEL8 options
 CONFIGURE_OPTS=(
     -Dsysvinit-path=/etc/rc.d/init.d
```

mrc0mmand:
@msekletar and another *sigh*: please add `timeout` to the aforementioned `dracut_install` as well to fix:

```
Jun 18 12:36:03 systemd-testsuite testsuite.sh[595]: /testsuite.sh: line 170: timeout: command not found
```

Let's hope it's the last missing dependency...

dtardon (approved):
LGTM

PR #112, opened 2020-07-01
(#1868877) mount-setup: fix segfault in mount_cgroup_controllers when using gcc9…

taskset:
… compiler

According to the documentation:
https://gcc.gnu.org/gcc-9/porting_to.html#complit

The 'join_controllers' that relied on the extended lifetime needs
to be fixed, move the compound literals to the function scope it
need to accessible in.


msekletar:
I've added BZ reference to the commit message and created new PR.

Closed in favour of #128

PR #113, opened 2020-08-06
(#1793533) Don't emit PropertiesChanged signal needlessly

msekletar:


jsynacek (approved):
LGTM.

gleventhal:
I'm not sure if you understand just how much impact this change really has for me, but just let me say thank you!  

For systems with busy automount activity, this behavior leads to SSH / session timeouts and CPU load as backlogged work and systemd/dbus/journald contend heavily for CPU when dbus gets exhausted (while iterating over >= ~3000 mounts).  

Thanks for fixing this!

PR #115, opened 2020-08-14
(#1868877) basic/virt: try the /proc/1/sched hack also for PID1

taskset:
If a container manager does not set $container, we could end up
in a strange situation when detect-virt returns container-other when
run as non-pid-1 and none when run as pid-1.

(cherry picked from commit 342bed02084c4396dd2f1054bd559bfb2699cfcb)
Resolves: #1868877

### DIFF
--- a/src/basic/virt.c
+++ b/src/basic/virt.c
@@ -427,7 +427,6 @@ finish:
 }
 
 int detect_container(void) {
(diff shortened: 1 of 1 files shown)


msekletar (approved):
LGTM

PR #120, opened 2020-09-29
(#1880270) logind: don't print warning when user@.service template is masked

dtardon:
User instance of systemd is optional feature and if user@.service
template is masked then administrator most likely doesn't want --user
instances of systemd for logged in users. We don't need to be verbose
about it.

(cherry picked from commit 03b6fa0c5b51b0d39334ff6ba183a3391443bcf6)

Resolves: #1880270


jsynacek (approved):
LGTM.

PR #125, opened 2020-10-21
(#1870638) socket: New option 'FlushPending' (boolean) to flush socket before en…

dtardon:
…tering listening state

Disabled by default. When Enabled, before listening on the socket, flush the content.
Applies when Accept=no only.

(cherry picked from commit 3e5f04bf6468fcb79c080f02b0eab08f258bff0c)

Resolves: #1870636


msekletar (approved):
LGTM

PR #126, opened 2020-10-22
(#1860374) seccomp: rework how the S[UG]ID filter is installed

dtardon:
If we know that a syscall is undefined on the given architecture, don't
even try to add it.

Try to install the filter even if some syscalls fail. Also use a helper
function to make the whole a bit less magic.

This allows the S[UG]ID test to pass on arm64.

(cherry picked from commit da4dc9a6748797e804b6bc92ad513d509abf581c)

Resolves: #1860374


msekletar (approved):
LGTM

PR #128, opened 2020-11-02
(#1868877) mount-setup: fix segfault in mount_cgroup_controllers when using gcc9

msekletar:
According to the documentation:
https://gcc.gnu.org/gcc-9/porting_to.html#complit

The 'join_controllers' that relied on the extended lifetime needs
to be fixed, move the compound literals to the function scope it
need to accessible in.

Resolves: #1868877


dtardon (approved):
LGTM

PR #131, opened 2020-11-04
(#1812972) allow instantiated units to be enabled via presets

dtardon:


src/shared/install.c
```diff
@@ -2868,15 +2916,71 @@ static int read_presets(UnitFileScope scope, const char *root_dir, Presets *pres
...
+                _cleanup_strv_free_ char **out_strv = NULL;
+                char **iter;
+
+                r = unit_name_to_prefix(unit_name, &prefix);
+                if (r < 0)
+                        return r;
+
+                STRV_FOREACH(iter, rule.instances) {
+                        _cleanup_free_ char *name = NULL;
+                        r = unit_name_build(prefix, *iter, ".service", &name);
```
sohankunkerkar: I think it would be helpful to add [this] (https://github.com/systemd/systemd/pull/16816) as well in this PR.

src/test/test-install-root.c
```diff
@@ -983,6 +983,63 @@ static void test_with_dropin_template(const char *root) {
...
+
+        assert_se(unit_file_get_state(UNIT_FILE_SYSTEM, root, "foo@.service", &state) >= 0 && state == UNIT_FILE_DISABLED);
+
+        p = strjoina(root, "/usr/lib/systemd/system-preset/test.preset");
+        assert_se(write_string_file(p,
+                                    "enable foo@.service bar0 bar1 bartest\n"
+                                    "enable emptylist@.service\n" /* This line ensures the old functionality for templated unit still works */
+                                    "disable *\n" , WRITE_STRING_FILE_CREATE) >= 0);
+
+        assert_se(unit_file_get_state(UNIT_FILE_SYSTEM, root, "foo@bar0.service", &state) >= 0 && state == UNIT_FILE_DISABLED);
```
sohankunkerkar, line 1007: could you please add a test case for non-service instantiated units also?

src/shared/install.c
dtardon: Yup, makes sense.

src/test/test-install-root.c
dtardon: Could you send a PR upstream?

msekletar (approved):
LGTM

PR #137, opened 2020-11-09
(#1876807) remove references of non-existent man pages

dtardon:
This is a follow-up to commit 8ad89170001c9aba8849630ddb5da81d9e24a1bc,
which introduced the man page change.

Resolves: #1876807


msekletar (approved):
LGTM

PR #139, opened 2020-11-10
(#1827462) udev/net_id: don't generate slot based names if multiple devices might claim the same slot

msekletar:


dtardon (approved):
LGTM

PR #140, opened 2020-11-26
(#1888912) reload SELinux label cache on daemon-reload

dtardon:


msekletar (approved):
LGTM

msekletar:
@mrc0mmand, please do make sure to run the internal test-suite (at least basic subset of it) also with SELinux disabled.

PR #141, opened 2020-12-04
(#1885143) bind_remount: avoid calling statvfs

dtardon:


src/basic/mount-util.c
```diff
@@ -313,55 +315,44 @@ int umount_recursive(const char *prefix, int flags) {
          * unmounting them until they are gone. */
 
         do {
-                _cleanup_fclose_ FILE *proc_self_mountinfo = NULL;
+                _cleanup_(mnt_free_tablep) struct libmnt_table *table = NULL;
+                _cleanup_(mnt_free_iterp) struct libmnt_iter *iter = NULL;
 
                 again = false;
                 r = 0;
```
msekletar: Nit, assigning to `r` seems redundant.

src/basic/mount-util.c
```diff
@@ -313,55 +315,44 @@ int umount_recursive(const char *prefix, int flags) {
...
+                        path = mnt_fs_get_target(fs);
+                        if (!path)
                                 continue;
 
-                        if (umount2(p, flags) < 0) {
-                                r = log_debug_errno(errno, "Failed to umount %s: %m", p);
+                        if (!path_startswith(path, prefix))
+                                continue;
+
+                        if (umount2(path, flags) < 0) {
```
msekletar: In upstream there was a follow-up (827ea5). It figure we should backport it too.

msekletar (changes requested):

src/basic/mount-util.c
dtardon: Indeed.

src/basic/mount-util.c
dtardon: Done.

msekletar (approved):
LGTM

PR #143, opened 2020-12-11
(#1657810) cgroup: Also set blkio.bfq.weight

dtardon:
Commit [1] added a workaround when unified cgroups are used but missed
legacy cgroups where there is the same issue.

[1] <https://github.com/systemd/systemd/commit/2dbc45aea747f25cc1c3848fded2ec0062f96bcf>

Signed-off-by: Pavel Hrdina <phrdina@redhat.com>
(cherry picked from commit 35e7a62ca32a30169a94693b831e53c832251984)

Resolves: #1657810


msekletar (approved):
LGTM

msekletar:
@dtardon you should probably rebase and force-push to retrigger CI

PR #145, opened 2021-01-06
(#1642728) udev: make algorithm that selects highest priority devlink less susceptible to race conditions

msekletar:


msekletar:
Travis CI failure is due to Dockerhub rate-limiting and hence inability of the setup script to set up CentOS 8 container on Ubuntu.

test/udev-test.pl
```diff
@@ -18,11 +18,19 @@
```
dtardon: Do we really need this commit?

dtardon: Well, it doesn't matter that much...

dtardon (approved):
LGTM

PR #150, opened 2021-01-26
(#1916835) seccomp: allow turning off of seccomp filtering via env var

msekletar:
Fixes: #17504

(While we are it, also move $SYSTEMD_SECCOMP_LOG= env var description
into the right document section)

Also suggested in: https://github.com/systemd/systemd/issues/17245#issuecomment-704773603

(cherry picked from commit ce8f6d478e3f6c6a313fb19615aa5029bb18f86d)

Resolves: #1916835


dtardon (approved):
LGTM

PR #161, opened 2021-03-10
(#1642460) backport UserStopDelaySec=

dtardon:


src/login/logind-user.c
```diff
@@ -385,68 +373,50 @@ int user_start(User *u) {
...
-                log_error("Failed to stop user service: %s", bus_error_message(&error, r));
-                return r;
-        }
+        u->service_job = mfree(u->service_job);
 
-        free_and_replace(u->service_job, job);
-        return r;
+        r = manager_stop_unit(u->manager, u->service, &error, &u->service_job);
+        if (r < 0)
+                log_warning_errno(r, "Failed to stop user service '%s', ignoring: %s", u->service, bus_error_message(&error, r));
```
msekletar, line 412: I think that error logging here should be symmetric with `user_start_service()`. Either make it both use  `log_full_errno()` or use `log_warning_errno()` in both cases.

msekletar (commented):
LGTM overall. The backport is quite big so thorough testing would be in order.

@mrc0mmand do we have/can we write any test case that would test UserStopDelaySec= and also user lingering?

mrc0mmand:
> @mrc0mmand do we have/can we write any test case that would test UserStopDelaySec= and also user lingering?

We don't, but I guess we should have. I'll look into it.

Also, needs rebase.

jamacku:
@dtardon @msekletar @mrc0mmand  Could we merge this one?

msekletar (approved):

PR #163, opened 2021-03-11
(#1934028) rc-local: order after network-online.target

dtardon:
I think this was the intent of commit 91b684c7300879a8d2006038f7d9185d92c3c3bf,
just network-online.target didn't exist back then.

RHEL-only

Resolves: #1934028


msekletar (approved):
LGTM

PR #175, opened 2021-05-17
(#1942299) Revert multipath symlink race fix

msekletar:


mrc0mmand:
I'd say the test fail is related, since it failed two times in a row and when trying it on master the issue is not present.

msekletar:
Hopefully I've fixed it up and CI will be green now.

dtardon (approved):
LGTM

PR #178, opened 2021-05-24
(#1927290) cgroup: Also set io.bfq.weight

jacekmigacz:
Current kernels with BFQ scheduler do not yet set their IO weight
through "io.weight" but through "io.bfq.weight" (using a slightly
different interface supporting only default weights, not per-device
weights). This commit enables "IOWeight=" to just to that.

This patch may be dropped at some time later.

Github-Link: https://github.com/systemd/systemd/issues/7057
Signed-off-by: Kai Krakow <kai@kaishome.de>

(cherry picked from commit 21221ce1ce9a572e82d46d80692afd65c224fc50)

Related: #1927290


jacekmigacz:
I've skipped merging NEWS intentionally, since it refers to v243.

dtardon (approved):
LGTM

PR #179, opened 2021-05-25
(#1946453) Backport upstream commit related to user-runtime-dir@.service

jacekmigacz:
user-runtime-dir: downgrade a few log messages to LOG_DEBUG that we ignore
units: make sure user-runtime-dir@.service is Type=oneshot
units: order user-runtime-dir@.service after systemd-user-sessions.service
units: assign user-runtime-dir@.service to user-%i.slice

bz1946453


dtardon (approved):
LGTM

PR #183, opened 2021-06-11
(#1959339) Add support for lenient initramfs detection

jacekmigacz:
util: rework in_initrd() to make use of path_is_temporary_fs()
initrd: do a debug log if /etc/initrd-release doesn't take effect
initrd: do a debug log if failed to detect rootfs type
initrd: do a debug log if /etc/initrd-release doesn't take effect

### DIFF
--- a/doc/ENVIRONMENT.md
+++ b/doc/ENVIRONMENT.md
@@ -37,6 +37,14 @@ All tools:
   useful for debugging, in order to test generators and other code against
   specific kernel command lines.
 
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Both current CI issues (stuck CentOS CI & `test-copy` fail under ASan/UBSan) should be fixed once https://github.com/redhat-plumbers/systemd-rhel8/pull/184 is merged.

msekletar (approved):
LGTM

PR #186, opened 2021-06-18
(#1954802) core: Hide "Deactivated successfully" message

jamacku:
Show message "Deactivated successfully" in debug mode (when manager is
user) rather than in info mode. This message has low information value
for regular users and it might be a bit overwhelming on a system with
a lot of devices.

(cherry picked from commit edf2ee22f54005d76b2fb8fdcc9c60974feb88bc)

Resolves: #1954802


dtardon (approved):
LGTM

PR #188, opened 2021-06-18
(#1968647) CVE-2019-3842 systemd: Spoofing of XDG_SEAT allows for actions to be checked against "allow_active" instead of "allow_any" [rhel-8.2.0.z]

jacekmigacz:
pam_systemd: simplify how we process env vars
pam-systemd: use secure_getenv() rather than getenv()

It is also applied for **XDG_VTNR**, **XDG_SESSION_TYPE**, **XDG_SESSION_CLASS** and **XDG_SESSION_DESKTOP** to make it consistent.

### DIFF
--- a/man/meson.build
+++ b/man/meson.build
@@ -68,7 +68,6 @@ foreach tuple : xsltproc.found() ? manpages : []
                 foreach htmlalias : htmlaliases
                         link = custom_target(
                                 htmlalias,
-                                input : p2,
                                 output : htmlalias,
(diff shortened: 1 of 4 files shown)


mrc0mmand:
The CentOS CI fail is expected, 8.2.0 is missing https://github.com/redhat-plumbers/systemd-rhel8/pull/107, not worth backporting.

dtardon (approved):
LGTM

PR #191, opened 2021-06-25
(#1970860) remove a left-over break

dtardon:
By the "same logic as above...", we want to continue to fallback here,
but the break prohibits that.

This is a follow-up for ee1aa61c4710ae567a2b844e0f0bb8cb0456ab8c .

(cherry picked from commit 99df1cb6f50875db513a5b45f18191460a150f3d)

Related: #1970860


mrc0mmand (approved):
Whoops, thanks for catching this, LGTM!

PR #194, opened 2021-07-05
(#1269726) extend inhibitor lock enforcement to root

jacekmigacz:
Some basic checks below:
```
# systemd-inhibit --what=shutdown --mode=block sleep infinity &
[1] 941
```
```
# systemctl --check-inhibitors=yes reboot
Operation inhibited by "sleep infinity" (PID 941 "systemd-inhibit", user root), reason is "Unknown reason".
User vagrant is logged in on sshd.
Please retry operation after closing inhibitors and logging out other users.
Alternatively, ignore inhibitors and users with 'systemctl reboot -i'.
```

With all these changes from `SD_BUS_METHOD` to `SD_BUS_METHOD_WITH_NAMES`; also `SD_BUS_SIGNAL` has become `SD_BUS_SIGNAL_WITH_NAMES`. It was not obligatory, but it is more consistent now.

### DIFF
--- a/src/login/logind-dbus.c
+++ b/src/login/logind-dbus.c
@@ -1698,14 +1698,14 @@ static int verify_shutdown_creds(
                 Manager *m,
                 sd_bus_message *message,
                 InhibitWhat w,
-                bool interactive,
                 const char *action,
                 const char *action_multiple_sessions,
                 const char *action_ignore_inhibit,
+                uint64_t flags,
                 sd_bus_error *error) {
 
         _cleanup_(sd_bus_creds_unrefp) sd_bus_creds *creds = NULL;
-        bool multiple_sessions, blocked;
+        bool multiple_sessions, blocked, interactive;
         uid_t uid;
         int r;
 
@@ -1728,6 +1728,7 @@ static int verify_shutdown_creds(
 
         multiple_sessions = r > 0;
         blocked = manager_is_inhibited(m, w, INHIBIT_BLOCK, NULL, false, true, uid, NULL);
+        interactive = flags & SD_LOGIND_INTERACTIVE;
 
         if (multiple_sessions && action_multiple_sessions) {
                 r = bus_verify_polkit_async(message, CAP_SYS_BOOT, action_multiple_sessions, NULL, interactive, UID_INVALID, &m->polkit_registry, error);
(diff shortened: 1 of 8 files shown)


mrc0mmand:
Don't mind the GCC job fails, there's something wrong with the API atm, and it's trying to checkout a different PR (which is already merged). I'll check if there's something I can do.

mrc0mmand:
@jacekmigacz could you, please, force-push this PR again? Just `git commit --am --no-ed && git push -f origin bz1269726-extend-inhibitor-lock-enforcement-to-root` should do the trick. For some reason GH actions have wrong metadata and keep trying to fetch the original PR (#193) instead of this one.

jamacku (commented):
I think this PR should be open against the ``master`` branch. We usually open PR against specific branches only when we are dealing with Z-stream updates.

mrc0mmand:
Yeah, I was just about to ask if we've already moved to the z-stream branches with the updates. The CI fails because the 8.5.0 branch lacks the necessary patches. I'll change the base of this PR to `master`, so @jacekmigacz doesn't have to open a new one.

src/login/logind-dbus.c
```diff
@@ -1765,20 +1773,32 @@ static int method_do_shutdown_or_sleep(
...
+                r = sd_bus_message_read(message, "b", &interactive);
+
         if (r < 0)
                 return r;
 
+        if (with_flags && (flags & ~SD_LOGIND_SHUTDOWN_AND_SLEEP_FLAGS_PUBLIC))
+                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS,
+                                         "Invalid flags parameter");
+
+        SET_FLAG(flags, SD_LOGIND_INTERACTIVE, interactive);
```
dtardon: Hmm, this seems wrong to me. If `with_flags == true`, then `interactive == false` here, so the flag is being unconditionally reset... Or am I missing something?

dtardon: I guess it wouldn't hurt to backport https://github.com/systemd/systemd/pull/18436 as well--that fixes the problem, if there indeed is one...

dtardon (approved):
LGTM

PR #205, opened 2021-08-03
(#1905582) Set default core ulimit to 0, but keep the hard limit ulimited

mrc0mmand:
so users can change it later.

Follow-up to 830bd662276ee117e65a4b3d541f77e8b172eafd.

rhel-only
Resolves: #1905582

### DIFF
--- a/src/core/system.conf.in
+++ b/src/core/system.conf.in
@@ -52,7 +52,7 @@
 #DefaultLimitFSIZE=
 #DefaultLimitDATA=
 #DefaultLimitSTACK=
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Verified locally:

```
# systemctl show | grep CORE
DefaultLimitCORE=infinity
DefaultLimitCORESoft=0
# ulimit -c
0
# sudo -s  -u test
$ ulimit -c
0
$ ulimit -c unlimited
$ ulimit -c
unlimited
```

dtardon (approved):
LGTM

PR #213, opened 2021-09-24
(#1860899) Avoid a race when exiting user service

dtardon:


msekletar (approved):
LGTM

jamacku:
@dtardon Could you please rebase. Thanks

jamacku:
@mrc0mmand Could you please have a look at CI. Thanks

mrc0mmand:
@lnykryn interesting, the GH API (as well as GH in general) had an outage and your bot just assumed that all CIs passed ;)

jamacku:
@mrc0mmand That's known issue. We should check both [check-runs](https://docs.github.com/en/rest/reference/checks#list-check-runs-for-a-git-reference) and [check-suites](https://docs.github.com/en/rest/reference/checks#list-check-suites-for-a-git-reference), but we check only ``check-suites``. I'm looking to it.

mrc0mmand:
> @mrc0mmand That's known issue. We should check both [check-runs](https://docs.github.com/en/rest/reference/checks#list-check-runs-for-a-git-reference) and [check-suites](https://docs.github.com/en/rest/reference/checks#list-check-suites-for-a-git-reference), but we check only `check-suites`. I'm looking to it.

I see, thank you. In the meantime, I fixed the CI issue in https://github.com/systemd/systemd-centos-ci/pull/441, so the CIs should be green once again.

jamacku:
@mrc0mmand @systemd-rhel-bot  shouldn't be able to merge PR with failing CI for now. I protected all branches and CI is now required together with one review.

mrc0mmand:
> @mrc0mmand @systemd-rhel-bot shouldn't be able to merge PR with failing CI for now. I protected all branches and CI is now required together with one review.

Excellent, thanks!

PR #219, opened 2021-10-12
(#2004765) Remove unintended additions to systemd-analyze man page

dtardon:
These changes were introduced in commit
a2e00522971897909db2a81b4daf10e5700f453e .

Resolves: #2004765


mrc0mmand:
@dtardon could you also disable `libiptc` in the unit tests workflow?

https://github.com/redhat-plumbers/systemd-rhel8/blob/793dc4d9e32baba27eac1f37283a7485b0889803/.github/workflows/unit_tests.sh#L53

We did that in the spec file, but the CI configuration still tries to build with it (and fails to do so).

mrc0mmand (approved):
LGTM

PR #220, opened 2021-10-14
test: Disable test-cap-list

jamacku:
Related: #1968648,#1968651

> Let's make sure that we are in sync with dist-git


mrc0mmand (approved):
LGTM

mrc0mmand:
Oh well, this breaks the compilation with `-Werror`, but let's hope there won't be any more 8.3 z-streams.

PR #222, opened 2021-10-15
(#1931806) journal: refresh cached credentials of stdout streams

dtardon:
journald assumes that getsockopt(SO_PEERCRED) correctly identifies the
process on the remote end of the socket. However, this is incorrect
according to man 7 socket:

    The returned  credentials  are  those that were in effect at the
    time of the call to connect(2) or socketpair(2).

This becomes a problem when a new process inherits the stdout stream
from a parent. First, log messages from the child process will
be attributed to the parent. Second, the struct ucred used by journald
becomes invalid as soon as the parent exits. Further sendmsg calls then
fail with ENOENT. Logs for the child process then vanish from the journal.

Fix this by using recvmsg on the stdout stream, and refreshing the cached
struct ucred if SCM_CREDENTIALS indicate a new process.

Fixes #13708

(cherry picked from commit 09d0b46ab61bebafe5bdc1be95ee153dfb13d6bc)

Resolves: #1931806

### DIFF
--- a/src/journal/journald-stream.c
+++ b/src/journal/journald-stream.c
@@ -484,10 +484,21 @@ static int stdout_stream_scan(StdoutStream *s, bool force_flush) {
 }
 
 static int stdout_stream_process(sd_event_source *es, int fd, uint32_t revents, void *userdata) {
+        uint8_t buf[CMSG_SPACE(sizeof(struct ucred))];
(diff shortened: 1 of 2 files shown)


msekletar (approved):
LGTM

PR #224, opened 2021-10-21
(#1991834) udev rules: add rule to create /dev/ptp_hyperv

dtardon:
As for the KVM case, necessary for network cards with
PTP devices when running a guest on HyperV

(cherry picked from commit 32e868f058da8b90add00b2958c516241c532b70)

Resolves: #1991834


msekletar (approved):
LGTM

PR #227, opened 2021-10-25
(#2017033) Disable libpitc to fix CentOS Stream CI

dtardon:
We have disabled it in our spec starting with 8.5.0, so let's follow
suit here.

Related: #2017033


mrc0mmand (approved):
LGTM

PR #230, opened 2021-11-02
(#2008825) Add implicit unit dependencies even if mount unit is generated from /proc/self/mountinfo

dtardon:


src/core/mount.c
```diff
@@ -1532,7 +1550,7 @@ static int mount_setup_existing_unit(
         }
 
         if (load_extras)
-                return mount_add_extras(MOUNT(u));
+                return mount_add_non_exec_dependencies(MOUNT(u));
```
msekletar, line 1553: Hmm, `mount_add_extra()` does a bit more than `mount_add_non_exec_dependendencies()`. Are we sure we want to change this?

dtardon: I think the change is OK. IIUC this `mount_add_extras` call serves a single purpose: to create proper dependencies for a network mount (in case the mount point has been remounted with `_netdev` option). OTOH, this commit is only needed for its "side effect" (the `mount_add_non_exec_dependencies` function), so I can revert this line...

msekletar: I think that your assumption is easily testable. In both cases mount some filesystem manually for the first time and if you are correct then you should end up with exactly the same mount unit, i.e. text dump for that unit should produce be the same in both cases. Please verify this holds before we merge this change as proposed.

msekletar (changes requested):

src/core/mount.c
dtardon: This indeed only adds remote-fs-related dependencies. The resulting unit is the same (modulo timestamp changes) with and without this patch.

msekletar: Thank you for double checking!

msekletar (approved):
LGTM

PR #232, opened 2021-11-04
(#2020239) Avoid possible assert on D-Bus call

dtardon:


msekletar (approved):
LGTM 

However, I wonder whether the change from @yuwata makes any difference.

https://github.com/systemd/systemd/pull/20935#issuecomment-974120315

PR #233, opened 2021-11-23
(#2017033) meson: do not fail if rsync is not installed with meson 0.57.2

jamacku:
https://github.com/mesonbuild/meson/issues/8641

Our CI started to fail. Even if the change is reverted in meson,
we need a quick workaround here.

(cherry picked from commit 7c5fd25119a495009ea62f79e5daec34cc464628)

Related: #2017033

### DIFF
--- a/man/meson.build
+++ b/man/meson.build
@@ -178,17 +178,20 @@ html = custom_target(
         depends : html_pages,
         command : ['echo'])
 
-run_target(
-        'doc-sync',
-        depends : man_pages + html_pages,
-        command : ['rsync', '-rlv',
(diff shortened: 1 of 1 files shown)


jamacku:
@mrc0mmand Could you please have a look at CI.

mrc0mmand:
I fixed the CentOS CI job. As for GH Actions, the main culprit is test-execute (which is "just" flaky and is missing patches from upstream) and test-seccomp (which strangely fails only on C8, not C8S). Anyway, both fails are unrelated.

dtardon (approved):
LGTM

PR #236, opened 2021-11-29
(#1860899) core: return true from cg_is_empty* on ENOENT

mrc0mmand:
(cherry picked from commit 1bcf3fc6c57d92927b96cad8c739099b4ceae236)

Related: #1860899

---

Forgotten(?) commit from https://github.com/systemd/systemd/pull/10428/ (follow-up to #213). Without this commit some tests (and most likely even machines) get stuck during shutdown with unified cgroups.

@dtardon @jamacku PTAL.

### DIFF
--- a/src/basic/cgroup-util.c
+++ b/src/basic/cgroup-util.c
@@ -1177,7 +1177,7 @@ int cg_is_empty(const char *controller, const char *path) {
 
         r = cg_enumerate_processes(controller, path, &f);
         if (r == -ENOENT)
-                return 1;
+                return true;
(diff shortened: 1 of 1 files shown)


mrc0mmand:
The GH Actions fail is unrelated, but I'll prep a PR for that as well.

mrc0mmand:
CentOS CI passed (which pointed out the original issue in the first place), so this PR should be ready to go.

dtardon (approved):
I wonder how did I manage to skip it... LGTM

PR #234, opened 2021-11-23
(#2017033) CI related backports

mrc0mmand:
Addresses outstanding CI issues (mentioned in #233).

PR #237, opened 2021-11-29
(#2017033) More CI fixes

mrc0mmand:
In #234 I *accidentally* backported bunch of CI-related code which 1) was ignored by gcc and 2) didn't even compile under gcc. Let's fix that (and the spuriously failing GH Actions as well).

### DIFF
--- a/src/basic/string-util.h
+++ b/src/basic/string-util.h
@@ -32,6 +32,12 @@ static inline bool streq_ptr(const char *a, const char *b) {
         return strcmp_ptr(a, b) == 0;
 }
 
+static inline char* strstr_ptr(const char *haystack, const char *needle) {
+        if (!haystack || !needle)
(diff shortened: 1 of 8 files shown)


mrc0mmand:
The CentOS CI fail should be fixed by #236.

src/basic/string-util.h
```diff
@@ -32,6 +32,12 @@ static inline bool streq_ptr(const char *a, const char *b) {
         return strcmp_ptr(a, b) == 0;
 }
 
+static inline char* strstr_ptr(const char *haystack, const char *needle) {
```
dtardon, line 35: This function is not used at the moment. OTOH, we might need it for future CI fixes--it's only used in tests so far--so let's get it in...

dtardon (approved):
LGTM

PR #136, opened 2020-11-06
(#1860899) avoid a race when exiting user service

dtardon:

PR #240, opened 2021-12-02
(#2024903) core: return true from cg_is_empty* on ENOENT

jamacku:
(cherry picked from commit 1bcf3fc6c57d92927b96cad8c739099b4ceae236)

Related: #2024903

---

Forgotten(?) commit from https://github.com/systemd/systemd/pull/10428/ (follow-up to #136). Without this commit some tests (and most likely even machines) get stuck during shutdown with unified cgroups.


dtardon (approved):
LGTM

PR #241, opened 2021-12-06
(#2017033) Various LGTM improvements

mrc0mmand:
@msekletar as for the *cleanup* stuff, I wrote a bunch of tests along with the query - https://github.com/mrc0mmand/codeql-sandbox/blob/master/cpp/UninitializedVariableWithCleanup/test.c - so if you have any other scenarios on your mind (or anything else which would be useful to detect), please let me know.


mrc0mmand:
The last meson-related commit is crucial, since with the latest meson (in C8S) there's a newline in the fetched `PATH=`, which breaks several tests. Calling `.strip()` fixes this.

dtardon (approved):
LGTM

PR #245, opened 2022-01-04
(#2036608) /p/s/mountinfo reparsing ratelimiting followups

msekletar:


src/libsystemd/libsystemd.sym
```diff
@@ -579,3 +579,8 @@ global:
         sd_event_source_get_ratelimit;
         sd_event_source_is_ratelimited;
 } LIBSYSTEMD_239;
+
+LIBSYSTEMD_250 {
```
dtardon, line 583: It probably doesn't matter in practice, but I wonder if we should rename this symbol version to RHEL_8_6 or something like that...

msekletar: If we do that would binaries compiled on RHEL work on Fedora without recompilation (not that anyone is actually doing it in practice but still...)? Or in other words, is the symbol versioned with @@RHEL_8_6 resolvable on Fedora if there the same symbol is versioned with @@LIBSYSTEMD_250?

dtardon: Hm, you're right. I doubt this would work. Let's scratch that idea.

dtardon (approved):
LGTM

PR #246, opened 2022-01-06
(#2037807) sysctl: Allow unprivileged users to send ICMP echo requests 

jamacku:
Resolves: [#2037807](https://bugzilla.redhat.com/show_bug.cgi?id=2037807)


dtardon (approved):
LGTM

PR #247, opened 2022-01-10
(#2039326) mount: don't propagate errors from mount_setup_unit() further up

kolyshkin:
_This is a backport of https://github.com/redhat-plumbers/systemd-rhel8/pull/244 to rhel-8.4.0 branch. To fix CI for stream8, I have also added a backport of https://github.com/redhat-plumbers/systemd-rhel8/pull/233._

If we can't process a specific line in /proc/self/mountinfo we should
log about it (which we do), but this should not affect other lines, nor
further processing of mount units. Let's keep these failures local.

Fixes: #10874

Cherry picked from commit ba0d56f55f2073164799be714b5bd1aad94d059a.
Trivial conflict in src/core/mount.c, function mount_load_proc_self_mountinfo,
due to local commit ca634baa10e. Also, due to the same commit, int k
is no longer used and is thus removed.

Resolves: #2039082

PR #243, opened 2021-12-15
(#2017033) test: make TEST-47 less racy

mrc0mmand:
Based on:
  - 2e7090e94d0c8b31d418555ab2f6a9b75318f6a4
  - e00e2e0b50bbd120290572c8d1242703fb98b34e
  - 197298ff9fc930de450330095cc5b67d165d0801

Related: #2017033

PR #248, opened 2022-01-11
(#2039327) mount: don't propagate errors from mount_setup_unit() further up

jamacku:
If we can't process a specific line in /proc/self/mountinfo we should
log about it (which we do), but this should not affect other lines, nor
further processing of mount units. Let's keep these failures local.

Fixes: #10874

(cherry picked from commit ba0d56f55f2073164799be714b5bd1aad94d059a)

Resolves: #2039327

### DIFF
--- a/.github/workflows/unit_tests.sh
+++ b/.github/workflows/unit_tests.sh
@@ -131,7 +131,7 @@ for phase in "${PHASES[@]}"; do
             # Pull a Docker image and start a new container
             docker pull quay.io/centos/centos:$CENTOS_RELEASE
             info "Starting container $CONT_NAME"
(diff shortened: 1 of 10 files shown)


mrc0mmand:
@jamacku both this and #247 are missing https://github.com/redhat-plumbers/systemd-rhel8/pull/237 and https://github.com/redhat-plumbers/systemd-rhel8/pull/243 to make CI happy.

kolyshkin:
> @jamacku both this and #247 are missing #237 and #243 to make CI happy.

I've managed to make #247 CI pass (alas I did it before I found the above hint so it took me some time to figure out all the commits that need to be backported), so @jamacku you can just pick up all of those from #247.

msekletar (approved):
LGTM

PR #256, opened 2022-01-21
(#2037807) Allow unprivileged users to send ICMP echo requests

jamacku:
Upstream PR https://github.com/systemd/systemd/pull/13191

Follow-up to #246 

Resolves: #2037807

### DIFF
--- a/src/sysctl/sysctl.c
+++ b/src/sysctl/sysctl.c
@@ -183,7 +183,7 @@ static int parse_file(OrderedHashmap *sysctl_options, const char *path, bool ign
 
                 existing = ordered_hashmap_get(sysctl_options, p);
                 if (existing) {
-                        if (streq(value, existing->value)) {
+                        if (streq_ptr(value, existing->value)) {
                                 existing->ignore_failure = existing->ignore_failure || ignore_failure;
                                 continue;
                         }
--- a/src/sysctl/sysctl.c
+++ b/src/sysctl/sysctl.c
@@ -26,25 +26,71 @@ static char **arg_prefixes = NULL;
 static bool arg_cat_config = false;
 static bool arg_no_pager = false;
 
+typedef struct Option {
+        char *key;
+        char *value;
+        bool ignore_failure;
+} Option;
+
+static Option *option_free(Option *o) {
+        if (!o)
+                return NULL;
+
+        free(o->key);
+        free(o->value);
+
+        return mfree(o);
+}
+
+DEFINE_TRIVIAL_CLEANUP_FUNC(Option*, option_free);
+DEFINE_HASH_OPS_WITH_VALUE_DESTRUCTOR(option_hash_ops, char, string_hash_func, string_compare_func, Option, option_free);
+
+static Option *option_new(
+                const char *key,
+                const char *value,
+                bool ignore_failure) {
+
+        _cleanup_(option_freep) Option *o = NULL;
+
+        assert(key);
+        assert(value);
+
+        o = new(Option, 1);
+        if (!o)
+                return NULL;
+
+        *o = (Option) {
+                .key = strdup(key),
+                .value = strdup(value),
+                .ignore_failure = ignore_failure,
+        };
+
+        if (!o->key || !o->value)
+                return NULL;
+
+        return TAKE_PTR(o);
+}
+
 static int apply_all(OrderedHashmap *sysctl_options) {
-        char *property, *value;
+        Option *option;
         Iterator i;
         int r = 0;
 
-        ORDERED_HASHMAP_FOREACH_KEY(value, property, sysctl_options, i) {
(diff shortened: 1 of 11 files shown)


src/sysctl/sysctl.c
```diff
@@ -26,25 +26,71 @@ static char **arg_prefixes = NULL;
...
+                return NULL;
+
+        free(o->key);
+        free(o->value);
+
+        return mfree(o);
+}
+
+DEFINE_TRIVIAL_CLEANUP_FUNC(Option*, option_free);
+DEFINE_HASH_OPS_WITH_VALUE_DESTRUCTOR(option_hash_ops, char, string_hash_func, string_compare_func, Option, option_free);
```
mrc0mmand, line 46: Oh well, looks like we don't even have the `HASH_OPS` destructor helpers in RHEL 8 (they were introduced in v240 by https://github.com/systemd/systemd/pull/10920). Given that's a quite large backport, I wonder what should be the next steps, since the `-` prefix is necessary there to make stuff work in certain containers (see https://github.com/systemd/systemd/issues/13177#issuecomment-514931461 for reference).

/cc @dtardon @msekletar

mrc0mmand:
CI seems to be happy, I gave `systemd-sysctl` a brief check with ASan+UBsan (since the original backport caused a memory corruption) and that's sorted out as well. Also, `ping` is now available to unprivileged users again.

Old version:
```
# sysctl -w net.ipv4.ping_group_range="1 0"
net.ipv4.ping_group_range = 1 0
# sudo -u test ping localhost
ping: socket: Operation not permitted
# /usr/lib/systemd/systemd-sysctl 
Couldn't write '0 2147483647' to '-net/ipv4/ping_group_range', ignoring: No such file or directory
# sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 1	0
# sudo -u test ping localhost
ping: socket: Operation not permitted
```

New version:
```
# sysctl -w net.ipv4.ping_group_range="1 0"
net.ipv4.ping_group_range = 1 0
# sudo -u test ping localhost
ping: socket: Operation not permitted
# /usr/lib/systemd/systemd-sysctl
# sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 0	2147483647
# sudo -u test ping localhost
PING localhost(localhost (::1)) 56 data bytes
64 bytes from localhost (::1): icmp_seq=1 ttl=64 time=0.046 ms
...
```

mrc0mmand:
And just for completeness, the "ignore" mechanic seems to work as well:

```
# echo "fs.protected_symlinks = 9999" >>/usr/lib/sysctl.d/50-default.conf
# /usr/lib/systemd/systemd-sysctl; echo $?
Couldn't write '9999' to 'fs/protected_symlinks': Invalid argument
1

# echo "-fs.protected_symlinks = 9999" >>/usr/lib/sysctl.d/50-default.conf
# /usr/lib/systemd/systemd-sysctl; echo $?
Couldn't write '9999' to 'fs/protected_symlinks', ignoring: Invalid argument
0
```

src/sysctl/sysctl.c
dtardon, line 46: This macro is defined in a following commit. That means that a build of this commit--e.g., during bisection--will fail. Please reorder the commits to avoid that.

dtardon (changes requested):

src/sysctl/sysctl.c
mrc0mmand: Sure, done.

msekletar (changes requested):
LGTM but before I approve...

@jamacku @mrc0mmand please triple-check there were no further follow-up bugfixes for these reworks. Hashmaps are in critical path and we must be 100% sure it is correct.

mrc0mmand:
> LGTM but before I approve...
> 
> @jamacku @mrc0mmand please triple-check there were no further follow-up bugfixes for these reworks. Hashmaps are in critical path and we must be 100% sure it is correct.

I went through the git log of each modified file:
   * `path_compare_func()` is not used anywhere and could be safely dropped (was done in 6906ac9a6920b1849db51450df7739100c5100ed), but that shouldn't matter
   * there's a memory leak in the path cache rebuilding... path, which is unrelated to this PR, but should be fixed nonetheless in the future (https://github.com/systemd/systemd/pull/15954/commits)
   * ~~there's c380b84d8b88ca61371d8fdb77d5719c778700fc (and follow-up ca3237150e965c33863addb9e89100030692d36d) which look like a good-to-have commits, so I pulled them in as well~~
     * nevermind, that would require a couple of other commits to be pulled in as well, not worth it imo
   * db99904bc8482efe556bb010a8b203a3e60ee37f is definitely something we should have; backported

That should be, hopefully, it.

mrc0mmand:
We should merge this soon, since more and more uses are hitting the issue - @msekletar PTAL.

msekletar (approved):
LGTM

PR #258, opened 2022-01-25
(#1795555) Provide a way for privileged users to interact with systemd user instances

jacekmigacz:


msekletar:
@jacekmigacz I see this is still marked as Draft. Please double-check what is still missing and drop "Draft" label if appropriate. Btw, also please rebase this work on top of main and force push to rerun all CIs.

PR #260, opened 2022-02-08
(#2017035) test: adapt to the new capsh format

mrc0mmand:
Since libcap v2.29 the format of cap_to_text() has been changed which
makes certain `test-execute` subtest fail. Let's remove the offending
part of the output (dropped capabilities) to make it compatible with
both the old and the new libcap.

(cherry picked from commit 9569e385036c05c0bf9fbccdbf3d131161398e2e)

Related: #2017035

### DIFF
--- a/test/test-execute/exec-capabilityboundingset-invert.service
+++ b/test/test-execute/exec-capabilityboundingset-invert.service
@@ -2,7 +2,7 @@
 Description=Test for CapabilityBoundingSet
 
 [Service]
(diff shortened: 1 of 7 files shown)


mrc0mmand:
The CentOS 8 CI fails are expected, as C8 went EOL and the content finally disappeared from mirrors as well. https://github.com/redhat-plumbers/systemd-rhel8/pull/261 should "fix it" by dropping the C8 job and keeping only the C8S ones.

jamacku (approved):
LGTM

PR #261, opened 2022-02-08
(#2017035) ci: drop CentOS 8 CI

mrc0mmand:
since it went EOL and we should use only Stream 8 from now on.

rhel-only
Related: #2017035


mrc0mmand:
The CentOS CI fail should be fixed by #260.

jamacku (approved):
LGTM

PR #262, opened 2022-02-15
(#2054264) seccomp: allow turning off of seccomp filtering via env var

dtardon:
Fixes: #17504

(While we are it, also move $SYSTEMD_SECCOMP_LOG= env var description
into the right document section)

Also suggested in: https://github.com/systemd/systemd/issues/17245#issuecomment-704773603

(cherry picked from commit ce8f6d478e3f6c6a313fb19615aa5029bb18f86d)

Resolves: #2054264

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: [centos8, stream8]
+        image: [stream8]
         phase: [GCC, GCC_ASAN]
     env:
       CONT_NAME: "systemd-centos8-ci"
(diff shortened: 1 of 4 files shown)


msekletar (approved):
LGTM

jamacku:
@mrc0mmand Could you please have a look at CI, thank you.

mrc0mmand:
RHEL 8.4 is apparently missing https://github.com/redhat-plumbers/systemd-rhel8/pull/260, but that can be backported later, since we're still about to do a couple of backports there anyway.

jamacku:
> RHEL 8.4 is apparently missing #260, but that can be backported later, since we're still about to do a couple of backports there anyway.

Thanks @mrc0mmand, I'm going to do a backport.

PR #263, opened 2022-02-23
(#2056527) core: unset HOME=/ that the kernel gives us

dtardon:
Partially fixes #12389.

%h would return "/" in a machine, but "/root" in a container. Let's fix
this by resetting $HOME to the expected value.

(cherry picked from commit 9d48671c62de133a2b9fe7c31e70c0ff8e68f2db)

Resolves: #2056527


msekletar (approved):
LGTM

PR #264, opened 2022-03-01
(#2051520) top-level drop-in support

jacekmigacz:


mrc0mmand:
After a quick glance I think we might want to backport 7a670b1dd981c645064f69faf85b04620aadbafb as well (didn't have time yet to test if it applies to RHEL 8 build as well, but given the code looks the same it probably does).

src/core/service.c
```diff
@@ -536,24 +536,9 @@ static int service_verify(Service *s) {
...
-                                return -ENOEXEC;
-                        }
-                        if (strv_isempty(command->argv)) {
-                                log_unit_error(UNIT(s),
-                                               "Service has an empty argv in %s=. Refusing.",
-                                               service_exec_command_to_string(c));
-                                return -ENOEXEC;
-                        }
-                }
+        if (!service_unit_name_is_valid(UNIT(s)->id)) {
```
dtardon: This hunk is wrong: it should only add a new check.

dtardon (changes requested):
The 1st commit erroneously removes existing code. Also, the 2nd and 3rd commits refer to invalid upstream commits.

src/core/service.c
jacekmigacz: Hunk remains in place, while that new check was removed in 3e1db806b0c18fd6138886ce67fac2655f09caef.

dtardon (approved):
LGTM

mrc0mmand (approved):
Also LGTM (and qa_ack'ed the respective BZ), thanks!

PR #265, opened 2022-03-03
(#2051981) journal-remote: check return value from MHD_add_response_header

dtardon:
Sadly, the API does not allow us to distinguish oom from invalid settings.
If the call fails, let's assume oom happened.

Coverity CID#1444714.

(cherry picked from commit 60d9c4f3b972ce70dfadc0a3f1a478a056c2ea7a)

Resolves: #2051981


msekletar (approved):
LGTM

PR #274, opened 2022-05-10
(#2039461) sd-bus: make BUS_DEFAULT_TIMEOUT configurable

dtardon:
This adds sd_bus_{get,set}_method_call_timeout().
If the timeout is not set or set to 0, then the timeout value is
parsed from $SYSTEMD_BUS_TIMEOUT= environment variable. If the
environment variable is not set, then built-in timeout is used.

(cherry picked from commit 385b2eb262a99373f09d01b7f5571dd71a14dc98)

Resolves: #2039461

### DIFF
--- a/doc/ENVIRONMENT.md
+++ b/doc/ENVIRONMENT.md
@@ -50,6 +50,11 @@ All tools:
   this only controls use of Unicode emoji glyphs, and has no effect on other
   Unicode glyphs.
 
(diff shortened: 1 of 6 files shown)


jamacku (commented):
Hmm, Bugzilla reference seems to point to wrong bug. @dtardon Is [#2049461](https://bugzilla.redhat.com/show_bug.cgi?id=2049461) correct?

dtardon:
Oops, fixed...

msekletar (approved):
LGTM

PR #281, opened 2022-05-28
(#1948480) unit: don't emit PropertiesChanged signal if adding a dependency to a…

jacekmigacz:
… unit is a no-op

(cherry picked from commit 5177cb0a9add4ae568cff6e6f7c2b3c77760c343)

Resolves: #1948480


dtardon (approved):
LGTM

PR #290, opened 2022-06-15
(#2095744) unit: check for mount rate limiting before checking active state

msekletar:
Having this check as part of mount_can_start() is too late because
UNIT(u)->can_start() virtual method is called after checking the active
state of unit in unit_start().

We need to hold off running mount start jobs when /p/s/mountinfo monitor
is rate limited even when given mount unit is already active.

Fixes #20329

(cherry picked from commit b161bc394b2cc8b271dda9208e310cc2af0cc29d)

Resolves: #2095744

### DIFF
--- a/src/core/mount.c
+++ b/src/core/mount.c
@@ -1999,9 +1999,6 @@ static int mount_can_start(Unit *u) {
 
         assert(m);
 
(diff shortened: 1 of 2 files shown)


dtardon (approved):
LGTM

PR #292, opened 2022-06-16
(#2095950) /p/s/mountinfo ratelimit followups

msekletar:


dtardon (commented):
Don't we also need https://github.com/redhat-plumbers/systemd-rhel8/pull/290?

msekletar:
We do but that is filed under different bugzilla so I will send separate PR for that.

dtardon:
> We do but that is filed under different bugzilla so I will send separate PR for that.

All right.

dtardon (approved):
LGTM

msekletar:
@dtardon  But before I submit that PR I think we will also need RHEL-8.4 variant of #280. Can you submit it once this is merged?

dtardon:
> @dtardon But before I submit that PR I think we will also need RHEL-8.4 variant of #280. Can you submit it once this is merged?

The code depends on this one. There is no point submitting it before this one is merged (and resolving merge conflicts two times).

msekletar:
> The code depends on this one. There is no point submitting it before this one is merged

Sure, that is why I wrote "...once this is merged" ;-)

dtardon:
Well, I didn't read past the first sentence :-)

https://github.com/redhat-plumbers/systemd-rhel8/pull/293

PR #295, opened 2022-06-22
(RHEL-9322) cgroup: drastically simplify caching of cgroups members mask

dtardon:
Previously we tried to be smart: when a new unit appeared and it only
added controllers to the cgroup mask we'd update the cached members mask
in all parents by ORing in the controller flags in their cached values.
Unfortunately this was quite broken, as we missed some conditions when
this cache had to be reset (for example, when a unit got unloaded),
moreover the optimization doesn't work when a controller is removed
anyway (as in that case there's no other way for the parent to iterate
though all children if any other, remaining child unit still needs it).
Hence, let's simplify the logic substantially: instead of updating the
cache on the right events (which we didn't get right), let's simply
invalidate the cache, and generate it lazily when we encounter it later.
This should actually result in better behaviour as we don't have to
calculate the new members mask for a whole subtree whever we have the
suspicion something changed, but can delay it to the point where we
actually need the members mask.

This allows us to simplify things quite a bit, which is good, since
validating this cache for correctness is hard enough.

Fixes: #9512
(cherry picked from commit 5af8805872809e6de4cc4d9495cb1a904772ab4e)

Resolves: [#2096371](https://bugzilla.redhat.com/show_bug.cgi?id=2096371)


src/core/unit.h
```diff
@@ -263,10 +263,10 @@ typedef struct Unit {
 
         /* Counterparts in the cgroup filesystem */
         char *cgroup_path;
-        CGroupMask cgroup_realized_mask;
-        CGroupMask cgroup_enabled_mask;
-        CGroupMask cgroup_subtree_mask;
-        CGroupMask cgroup_members_mask;
+        CGroupMask cgroup_realized_mask;           /* In which hierarchies does this unit's cgroup exist? (only relevant on cgroupsv1) */
+        CGroupMask cgroup_enabled_mask;            /* Which controllers are enabled (or more correctly: enabled for the children) for this unit's cgroup? (only relevant on cgroupsv2) */
+        CGroupMask cgroup_invalidated_mask;        /* A mask specifiying controllers which shall be considered invalidated, and require re-realization */
```
msekletar: `cgroup_invalidated_mask` isn't used anywhere. Also a comment contains typo in "specifiying".

msekletar (changes requested):
I think we shouldn't just cherry pick the single commit from the lengthy series in https://github.com/systemd/systemd/pull/10894 and hope for the best. I have a gut feeling this will introduce regressions in our handling of cgroups. At very least we shouldn't be introducing dead code.

src/core/unit.h
dtardon: I attempted to backport the whole series at first, but abandoned that approach in the end. The PR depends on previous PR(s). It would likely take several days to sort it out and backport everything needed.

jamacku:
@dtardon could you please rebase

Also please update commits to link to JIRA issue - [RHEL-9322](https://issues.redhat.com/browse/RHEL-9322)

msekletar:
CI failures looked related but to be sure they are not just flukes @lnykryn force-pushed the feature branch to rerun the checks.

jamacku:
Unit tests seem to be failing because of #434, But I'm not sure about CentOS CI.

msekletar:
CentOS CI failure is concerning. @dtardon can you please have a look?

mrc0mmand:
Oh well, the CentOS CI for this repo is currently FUBAR, since CentOS Stream 8 is no more. I'll need to figure something out to make it working again...

msekletar:
@mrc0mmand Maybe we could arrange "Redhat developer subscription for teams", i.e. self-service subscription and then we could generate activation key that we would store as GH secret and then use that to enable all needed repos in UBI RHEL container image (e.g. CRB).

mrc0mmand:
> @mrc0mmand Maybe we could arrange "Redhat developer subscription for teams", i.e. self-service subscription and then we could generate activation key that we would store as GH secret and then use that to enable all needed repos in UBI RHEL container image (e.g. CRB).

I mean, yeah, that would work for the container stuff (and I plan to actually do that), but this wouldn't work for CentOS CI, since, as the name suggests, it only supports CentOS. So I'll move the C8S job to C9S, which should work just fine. The environment will be different, but it should be enough for making sure we haven't royally screwed something up (we run the same tests internally on actual RHEL 8 anyway).

mrc0mmand:
CentOS CI should be back (albeit with some compromises, see https://github.com/systemd/systemd-centos-ci/commit/f415a75eadff46bf885d7e772cd63e342b707d96). I'll look into the container stuff next.

mrc0mmand:
A quick workaround for the current C8S containers is in https://github.com/redhat-plumbers/systemd-rhel8/pull/440.

msekletar:
@dtardon can you please force-push in order to retrigger CI, thank you!

msekletar:
CentOS CI failures are described by #434. However, test-execute failure in GHA is new to me. @mrc0mmand would be great if you could have a look.

msekletar (approved):
LGTM

mrc0mmand:
> CentOS CI failures are described by #434. However, test-execute failure in GHA is new to me. @mrc0mmand would be great if you could have a look.

The process in question died with `CLD_DUMPED`, but the test expects `CLD_KILLED`. This is somewhat expected and was addressed by systemd/systemd@c3ab2c389ee60d92fb8d7fe779ae9c4e3c092e4c and systemd/systemd@a429223d1767a9d6cb0a95fd2fa3b0d64ce3d4e7 by ignoring the difference between `CLD_DUMPED` and `CLD_KILLED` in `test-execute` (and also by disabling the pointless coredumps). Not sure why it popped up just now in the CI (most likely some kernel change, since the tests in GHA run in a C8S container on a Ubuntu 22.04 host).

PR #296, opened 2022-06-23
(#1857969) logind-session-dbus: allow to set display name via dbus

dtardon:
Currently, the only way to set display name of a graphical session is to
pass it to CreateSession(). But modern display managers like gdm start
the display server as part of the user session, which means that the
display name isn't known yet when the session is being created. Hence,
let's make it possible to set it later.

(cherry picked from commit 4885d7490b23e08d8444e5a68927ce9ce8727e5a)

Resolves: [#1857969](https://bugzilla.redhat.com/show_bug.cgi?id=1857969)


msekletar (approved):
LGTM

PR #299, opened 2022-07-18
(#1873540) Allow to switch journal back from /var to /run on shutdown

dtardon:
Resolves: [#1873540](https://bugzilla.redhat.com/show_bug.cgi?id=1873540)

### DIFF
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -47,6 +47,7 @@
 #include "log.h"
 #include "logs-show.h"
 #include "mkdir.h"
+#include "mount-util.h"
 #include "pager.h"
 #include "parse-util.h"
 #include "path-util.h"
@@ -162,6 +163,7 @@ static enum {
         ACTION_UPDATE_CATALOG,
         ACTION_LIST_BOOTS,
         ACTION_FLUSH,
+        ACTION_RELINQUISH_VAR,
         ACTION_SYNC,
         ACTION_ROTATE,
         ACTION_VACUUM,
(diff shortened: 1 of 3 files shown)


msekletar:
It would be quite risky to backport such a large PR at this point in development cycle of RHEL-8.7. I think we should either think about how to make this work w/o varlink or merge this early in the dev cycle backed by green test results, fuzzing, coverity scan review as this can't be realistically reviewed.

dtardon:
I don't think we should merge this as-is either. I did a big part of this--practically all the preparatory work concerning json and varlink support--long ago (in the 8.2-8.3 times), but then put it down and later forgot about it. I finished it out of curiosity, to see how big the change was going to be and if it's even going to pass CI...

If the functionality is desirable, I think this could be trimmed down neatly by using a signal, just like we do for `--sync` and `--flush`. OTOH, there's been no push on this from anyone, so maybe the best course is to do nothing at all...

dtardon:
I've reimplemented this by using a signal for the "relinquish" action. This trimmed it down to 3 commits, with only small divergences from upstream. It DTRT on both systems with a separate /var and without one.

PTAL.

msekletar (changes requested):
@dtardon this is much better. Thanks! 

However, before we merge this. I think we should try to propose the same change also upstream, i.e. propose PR that adds RTMIN+2 handling + document this behaviour. The fact this is not done is probably oversight.

dtardon:
> @dtardon this is much better. Thanks!
> 
> However, before we merge this. I think we should try to propose the same change also upstream, i.e. propose PR that adds RTMIN+2 handling + document this behaviour. The fact this is not done is probably oversight.

I can do that. However, I'm not sure it's really needed. I assume signals were originally used just because it was the easiest way to let `journalctl` send the necessary commands to `systemd-journald`. Now that we have varlink, the signals are no longer needed for that purpose, hence there's no need to add new ones. But maybe I'm wrong...

In this PR, I consider the use of a signal an impl. detail. That's why I haven't documented it.

dtardon:
> However, before we merge this. I think we should try to propose the same change also upstream, i.e. propose PR that adds RTMIN+2 handling + document this behaviour. The fact this is not done is probably oversight.

https://github.com/systemd/systemd/pull/26442

dtardon:
> [systemd/systemd#26442](https://github.com/systemd/systemd/pull/26442)

... and it's been rejected by Lennart.

msekletar (approved):
LGTM

msekletar:
> > [systemd/systemd#26442](https://github.com/systemd/systemd/pull/26442)
> 
> 
> 
> ... and it's been rejected by Lennart.

Thank you regardless!

PR #301, opened 2022-07-19
(#2087152) A couple of CI fixes for recent failures

mrc0mmand:


mrc0mmand:
@jamacku, @dtardon PTAL, this should unblock recent PRs like #300 and #299, where GH Actions get stuck and eventually timeout.

jamacku (approved):
LGTM

PR #310, opened 2022-08-18
(#2087152) test-procfs-util: skip test on certain errors

lnykryn:
Inspired by upstream bf47f71c1c

RHEL-only

### DIFF
--- a/src/test/test-procfs-util.c
+++ b/src/test/test-procfs-util.c
@@ -7,6 +7,7 @@
 #include "procfs-util.h"
 #include "process-util.h"
 #include "util.h"
+#include "tests.h"
 
 int main(int argc, char *argv[]) {
         char buf[CONST_MAX(FORMAT_TIMESPAN_MAX, FORMAT_BYTES_MAX)];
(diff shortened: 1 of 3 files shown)


src/test/test-procfs-util.c
```diff
@@ -52,8 +52,9 @@ int main(int argc, char *argv[]) {
                 log_info("Reducing limit by one to %"PRIu64"…", v-1);
 
                 r = procfs_tasks_set_limit(v-1);
-                log_info_errno(r, "procfs_tasks_set_limit: %m");
-                assert_se(r >= 0 || ERRNO_IS_PRIVILEGE(r) || r == -EROFS);
+                if (IN_SET(r, -ENOENT, -EROFS) || ERRNO_IS_PRIVILEGE(r))
+                        return log_tests_skipped_errno(r, "can't set tasks limit");
```
mrc0mmand, line 57: We don't have 317bb217d3f21312e58d9efdc6d3739f7895a08c in RHEL 8, so the easiest way to fix this would be:

```suggestion
                        return EXIT_TEST_SKIP;
```

mrc0mmand: Or by backporting the mentioned commit, if you're brave enough ;)

jamacku (approved):
LGTM

PR #311, opened 2022-08-23
(#2068575) scope: allow unprivileged delegation on scopes

msekletar:
Previously it was possible to set delegate property for scope, but you
were not able to allow unprivileged process to manage the scope's cgroup
hierarchy. This is useful when launching manager process that  will run
unprivileged but is supposed to manage its own (scope) sub-hierarchy.

Fixes #21683

(cherry picked from commit 03860190fefce8bbea3a6f0e77919b882ade517c)

Resolves: #2068575


dtardon (approved):
LGTM

PR #315, opened 2022-08-26
(#2110549) resolved: pin stream while calling callbacks for it

jamacku:
These callbacks might unref the stream, but we still have to access it,
let's hence ref it explicitly.

Maybe fixes: #10725

(cherry picked from commit d973d94dec349fb676fdd844f6fe2ada3538f27c)

Resolves: #2110549

### DIFF
--- a/src/resolve/resolved-dns-stream.c
+++ b/src/resolve/resolved-dns-stream.c
@@ -42,6 +42,8 @@ static int dns_stream_update_io(DnsStream *s) {
 }
 
 static int dns_stream_complete(DnsStream *s, int error) {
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Well, *somebody* ignored CI results in https://github.com/redhat-plumbers/systemd-rhel8/pull/311 where TEST-19 also fails, and where the relevant change was introduced, because it's missing `useradd` (and most likely `userdel` as well). Adding them to

https://github.com/redhat-plumbers/systemd-rhel8/blob/5b5571de21d1ddf9a00511a6b2f25d630a903f05/test/test-functions#L26

should, hopefully, make the test happy.

msekletar:
@mrc0mmand good catch, thanks!

msekletar (approved):
LGTM

PR #318, opened 2022-08-30
(#2110546) resolved: pin stream while calling callbacks for it

jamacku:
These callbacks might unref the stream, but we still have to access it,
let's hence ref it explicitly.

Maybe fixes: #10725

(cherry picked from commit d973d94dec349fb676fdd844f6fe2ada3538f27c)

Resolves: #2110546


dtardon (approved):
LGTM

PR #323, opened 2022-09-05
(#2123801) Add trigger limit for path units

dtardon:
Resolves: [#2123801](https://bugzilla.redhat.com/show_bug.cgi?id=2123801)


msekletar (approved):
LGTM

PR #327, opened 2022-09-19
(#2127170) core: move reset_arguments() to the end of main's finish

dtardon:
Fixes #16991

fb39af4ce42d7ef9af63009f271f404038703704 replaced `free_arguments()` with `reset_arguments()`, which frees arg_* variables as before, but also resets all of them to the default values. `reset_arguments()` was positioned in such a way that it overrode some arg_* values still in use at shutdown.

To avoid further unintentional resets, I moved `reset_arguments()` right before the return, when nothing else will be using the arg_* variables.

(cherry picked from commit 7d9eea2bd3d4f83668c7a78754d201b226acbf1e)

Resolves: [#2127170](https://bugzilla.redhat.com/show_bug.cgi?id=2127170)

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -2631,7 +2631,6 @@ finish:
                 m = manager_free(m);
         }
 
-        reset_arguments();
         mac_selinux_finish();
 
         if (reexecute)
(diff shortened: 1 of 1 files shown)


msekletar (approved):
LGTM

jamacku:
The patch was manually reverted, and PR was resubmitted as:

- #334

PR #341, opened 2022-11-18
(#2143101) basic: recognize pdfs filesystem as a network filesystem

msekletar:
Fujitsu advises their users to always use _netdev mount option with pdfs mounts. Hence it makes sense to simply consider pdfs mounts as network filesystem mounts.

https://software.fujitsu.com/jp/manual/manualfiles/m130027/j2ul1563/02enz200/j1563-02-06-02-02.html

RHEL-only

Resolves: #2143101


dtardon (approved):
LGTM

PR #345, opened 2022-12-08
(#2136738) Use target process context to set socket context

dtardon:
Note: The first commit is not strictly needed, but it's closely related.

Resolves: [#2136738](https://bugzilla.redhat.com/show_bug.cgi?id=2136738)


msekletar (approved):
LGTM

PR #346, opened 2022-12-08
(#1866955) Remember our idle state and use it to detect idle level transitions

dtardon:
Resolves: [#1866955](https://bugzilla.redhat.com/show_bug.cgi?id=1866955)


msekletar (changes requested):
@dtardon it seems that CI failure is related, in log I can see that `testsuite.sh` script was not found. Probably `setup_testsuite` function works differently in RHEL-8 and upstream.

dtardon:
Whew, this test is racy as hell! I've had to increase one of the sleeps and add an additional one--probably unnecessarily large, but I don't care--to make it pass.

msekletar (approved):
LGTM

PR #348, opened 2023-01-04
(#2158160) sysctl: downgrade message when we have no permission

msekletar:
We need to run sysctl also in containers, because the network subtree is namespaces and may legitimately be writable. But logging all "errors" at notice level creates unwanted noise.

Also downgrade message about missing sysctls to log_info. This might also be relatively common when configuration is targeted at different kernel versions. With log_debug it'll still end up in the logs, but isn't really worth of "notice" most of the time.

https://bugzilla.redhat.com/show_bug.cgi?id=1609806 (cherry picked from commit 32458cc9687c1b60ff0f22c0e71da93ce78b1534)

Resolves: #2158160


jamacku (approved):
LGTM

PR #353, opened 2023-01-17
(#2158832) Backport systemd-pstore

dtardon:
Resolves: [#2158832](https://bugzilla.redhat.com/show_bug.cgi?id=2158832)


msekletar (changes requested):
Looks mostly OK. 

However, in upstream the pstore service is pulled from sysinit.target. Also there are no dependencies on systemd-remount-fs because those should be generated automatically due to StateDiretory= option, i.e. /var/lib/systemd/pstore needs to be mounted before. 

I think we shouldn't deviate unnecessarily because it creates confusion later on, hence I'd like to see those changes pulled in as well.

msekletar:
@mrc0mmand it would be great if we had a integration test for pstore functionality.

For example crashing VM with configured pstore, then rebooting it (both done via qemu commands) and checking that pstore entry was created and then successfully transferred to /var/lib/systemd/pstore.

Do you think it would be possible to have such test case upstream (probably yes, but I worry about reliability)?

dtardon:
> However, in upstream the pstore service is pulled from sysinit.target. Also there are no dependencies on systemd-remount-fs because those should be generated automatically due to StateDiretory= option, i.e. /var/lib/systemd/pstore needs to be mounted before.
> 
> I think we shouldn't deviate unnecessarily because it creates confusion later on, hence I'd like to see those changes pulled in as well.

Indeed. I forgot to look at changes to the unit file. Updated.

mrc0mmand:
> @mrc0mmand it would be great if we had a integration test for pstore functionality.
> 
> For example crashing VM with configured pstore, then rebooting it (both done via qemu commands) and checking that pstore entry was created and then successfully transferred to /var/lib/systemd/pstore.
> 
> Do you think it would be possible to have such test case upstream (probably yes, but I worry about reliability)?

That would make sense I guess. Could you, please, open an upstream issue about this, so I won't forget?

msekletar (approved):
LGTM

cgwalters:
This seems to have missed backporting https://github.com/systemd/systemd/commit/f00c36641a253f4ea659ec3def5d87ba1336eb3b
which is required for the service to work by default, because the service will fail if the directory doesn't exist.  On ostree (and other image based systems) it's expected for `/var` to start out empty, and to have skeleton directories populated either by tmpfiles.d or by the service itself.

PR #355, opened 2023-01-17
(#2053273) Let SysV compat. shutdown commands honor inhibitors for root

dtardon:
Resolves: [#2053273](https://bugzilla.redhat.com/show_bug.cgi?id=2053273)

### DIFF
--- a/src/systemctl/systemctl.c
+++ b/src/systemctl/systemctl.c
@@ -8655,6 +8655,65 @@ static int logind_schedule_shutdown(void) {
 #endif
 }
 
+static int halt_main_old(void) {
+        int r;
+
+        r = logind_check_inhibitors(arg_action);
+        if (r < 0)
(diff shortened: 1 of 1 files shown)


dtardon:
@msekletar It just occurred to me that there's an easier--and safer--way than trying to make the _new_ code backwards-compatible again. Given that the changes are in a single function `halt_main`, we could add the whole function under a new name and then call the either old one or the new one as needed. Something like:
```
static int halt_main(void) {
        if (env_var_is_set())
                return halt_main_new();
        else
                return halt_main_old();
}
```

msekletar:
That sounds much better. Please try to implement it this way so we know how exactly would resulting code look.

dtardon:
Done.

msekletar (approved):
LGTM

PR #358, opened 2023-01-26
(#2059633) core: bring manager_startup() and manager_reload() more inline

msekletar:
Both functions do partly the same, let's make sure they do it in the same order, and that we don't miss some calls.

This makes a number of changes:

1. Moves exec_runtime_vacuum() two calls down in manager_startup(). This should not have any effect but makes manager_startup() more like manager_reload().

2. Calls manager_recheck_journal(), manager_recheck_dbus(), manager_enqueue_sync_bus_names() in manager_startup() too. This is a good idea since during reeexec we pass through manager_startup() and hence can't assume dbus and journald weren't up yet, hence let's check if they are ready to be connected to.

3. Include manager_enumerate_perpetual() in manager_reload(), too. This is not strictly necessary, since these units are included in the serialization anyway, but it's still a nice thing, in particular as theoretically the deserialization could fail.

(cherry picked from commit 3ad2afb6a204513c7834c64ab864e40169874390)

Resolves: #2059633

### DIFF
--- a/src/core/manager.c
+++ b/src/core/manager.c
@@ -1660,11 +1660,11 @@ int manager_startup(Manager *m, FILE *serialization, FDSet *fds) {
         /* Release any dynamic users no longer referenced */
         dynamic_user_vacuum(m, true);
 
-        exec_runtime_vacuum(m);
-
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

PR #361, opened 2023-02-16
(#2170084) pam_systemd: suppress LOG_DEBUG log messages if debugging is off

dtardon:
In the PAM module we need to suppress LOG_DEBUG messages manually, if debug logging is not on, as PAM won't do this for us. We did this correctly for most log messages already, but two were missing. Let's fix those too.

Fixes: #10822
(cherry picked from commit 2675747f3cdd6f1e6236bbb2f79abfa53fb307f1)

Resolves: [#2170084](https://bugzilla.redhat.com/show_bug.cgi?id=2170084)


msekletar (approved):
LGTM

PR #363, opened 2023-02-17
(#2169932) Fix lost control of the current terminal when default.target is rescue.target

dtardon:
When default.target is rescue.target, exiting from the single-user shell results in lost of the control of the current terminal. This is because the operation performed to continue to boot is systemctl default but default.target is now rescue.target and it is already active. Hence, no new process that controls the current terminal is created. Users need to make hardware reset to recover the situation.

This sounds like a bit corner case issue and some might feel configuring default.target as rescue.target is odd because there are several other ways to transition to rescue.mode without configuring default.target to rescue.target such as systemctl rescue or systemd.unit=rescue.target something like that. However, users unfamiliar with systemd operations tend to come up with systemctl set-default rescue.target.

To fix this issue, let's transition to default.target only when default.target is inactive. Otherwise, invoke the single-user shell again to keep control of the current terminal for users.

This new logic depends on whether D-Bus working well. Exiting without any check of result of systemctl default could lead to again the control lost of the current terminal. Hence, add checking results of each D-Bus operations including systemctl default and invoke the single-user shell if they fail.

(cherry picked from commit 937ca8330d11e406b8ef343bead6f4f6244e39c7)

Resolves: [#2169932](https://bugzilla.redhat.com/show_bug.cgi?id=2169932)

### DIFF
--- a/src/sulogin-shell/sulogin-shell.c
+++ b/src/sulogin-shell/sulogin-shell.c
@@ -14,6 +14,8 @@
 #include "process-util.h"
 #include "sd-bus.h"
 #include "signal-util.h"
+#include "special.h"
+#include "unit-def.h"
 
 static int reload_manager(sd_bus *bus) {
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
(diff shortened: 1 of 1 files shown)


mrc0mmand:
@dtardon could you force-push this one? I restarted failing jobs in all 8.9 PRs after #374 was merged, but I can't do it in this one, since the CentOS CI Jenkins moved to a new location and this PR predates it.

msekletar (approved):
LGTM

PR #364, opened 2023-02-22
(#2156780) logind-session: make stopping of idle session visible to admins

dtardon:
(cherry picked from commit 6269ffe7ee8a659df7336a2582054ecd9eecf4b1)

Resolves: [#2156780](https://bugzilla.redhat.com/show_bug.cgi?id=2156780)

### DIFF
--- a/src/login/logind-session.c
+++ b/src/login/logind-session.c
@@ -673,7 +673,7 @@ static int session_dispatch_stop_on_idle(sd_event_source *source, uint64_t t, vo
 
         idle = session_get_idle_hint(s, &ts);
         if (idle) {
-                log_debug("Session \"%s\" of user \"%s\" is idle, stopping.", s->id, s->user->name);
+                log_info("Session \"%s\" of user \"%s\" is idle, stopping.", s->id, s->user->name);
 
                 return session_stop(s, /* force */ true);
         }


msekletar (approved):
LGTM

jamacku:
@mrc0mmand Is ci failure expected? It seems related. Thanks

mrc0mmand:
> @mrc0mmand Is ci failure expected? It seems related. Thanks

It's not related to this PR, the test is just extremely racy (see https://github.com/redhat-plumbers/systemd-rhel8/pull/346#issuecomment-1431885217). /cc @dtardon

dtardon:
It seems the test for #346 is not reliable enough after all. I really don't want to backport later improvements to it, as they depend on general improvements to `TEST-35-LOGIN`, so we probably should disable it.

dtardon:
> It seems the test for #346 is not reliable enough after all. I really don't want to backport later improvements to it, as they depend on general improvements to `TEST-35-LOGIN`, so we probably should disable it.

Or we could just pick up the necessary minimum that's needed to run the test from the current upstream version instead of backporting separate commits (which didn't even occur to me when I was doing the backport, because I'm too used to the latter). That shouldn't be that much work...

PR #365, opened 2023-03-13
(#2176892) Make sure journalctl --flush does not hang

dtardon:
This is not a change I'm proud of, but it fixes the regression and it is good enough for the original use case...

Resolves: [#2176892](https://bugzilla.redhat.com/show_bug.cgi?id=2176892)


msekletar:
CI failure is the same as in #366.

msekletar (approved):
LGTM

PR #366, opened 2023-03-13
(#2174645) Make sure journalctl --flush does not hang

dtardon:
This is not a change I'm proud of, but it fixes the regression and it is good enough for the original use case...

Resolves: [#2174645](https://bugzilla.redhat.com/show_bug.cgi?id=2174645)


msekletar:
CI failure seems unrelated (although we should really work on better reliability of logind test case).

msekletar (approved):
LGTM

PR #370, opened 2023-03-15
Use less in "secure" mode when run under sudo

dtardon:
Resolves: [#2175623](https://bugzilla.redhat.com/show_bug.cgi?id=2175623)

### DIFF
--- a/man/less-variables.xml
+++ b/man/less-variables.xml
@@ -37,12 +37,30 @@
     </varlistentry>
 
     <varlistentry id='lesssecure'>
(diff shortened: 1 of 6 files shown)


msekletar:
CI failure is unrelated and already addressed in #374

msekletar (dismissed):
LGTM

msekletar:
Hi @dtardon, we need one followup because of how the build system is set up for RHEL-8. I've created PR #376 with the follow-up for `rhel-8.7.0` branch, please add that commit also here.

jamacku (approved):
LGTM

PR #371, opened 2023-03-15
Use less in "secure" mode when run under sudo

dtardon:
Resolves: [#2175622](https://bugzilla.redhat.com/show_bug.cgi?id=2175622)


msekletar (approved):
LGTM

PR #373, opened 2023-03-29
(#2180380) journal-vacuum: count size of all journal files

dtardon:
Currently, active journal files are excluded, which means that vacuuming may not remove anything even if *MaxUse= has been exceeded.

(cherry picked from commit 9ea46af4f2368b41d57705bac09774778126507f)

Resolves: [#2180380](https://bugzilla.redhat.com/show_bug.cgi?id=2180380)


msekletar (approved):
LGTM

PR #374, opened 2023-03-29
(#2179309) test: make TEST-35-LOGIN stable again

mrc0mmand:
Let's add a couple of synchronization points and reduce the amount of sleeps to make the test, hopefully, stable again.

Follow-up to 638c2418e7.
Related: #2179309
rhel-only

---

Let's see if the CI agrees, locally the test appeared quite stable once again.

### DIFF
--- a/test/TEST-35-LOGIN/testsuite.sh
+++ b/test/TEST-35-LOGIN/testsuite.sh
@@ -3,6 +3,141 @@
 set -eux
 set -o pipefail
 
+cleanup_test_user() (
+    set +ex
+
+    pkill -u "$(id -u logind-test-user)"
+    sleep 1
+    pkill -KILL -u "$(id -u logind-test-user)"
+    userdel -r logind-test-user
+
+    return 0
+)
+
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Welp, CI definitely disagrees. The search goes on then...

dtardon (dismissed):
Anything that makes the test more reliable is good.

LGTM

dtardon:
Btw, maybe you could take inspiration from https://github.com/systemd/systemd/pull/27038 and use a fifo for synchronization with the expect process instead of a busy loop?

mrc0mmand:
Okay, so I gave up on the expect stuff, since it did everything but what I'd expect, and decided to just backport the latest version of the test from the upstream. After tweaking it a little, to account for missing stuff in downstream, it seems to be pretty stable (at least it keeps passing locally when I run it in a loop). Let's see what the CI is going to say.

msekletar:
LGTM and CI is finally green 🥳

msekletar (approved):

PR #377, opened 2023-03-31
(#2175622) test: ignore ENOMEDIUM error from sd_pid_get_cgroup()

msekletar:
Ubuntu builds on the Launchpad infrastructure run inside a chroot that does not have the sysfs cgroup dirs mounted, so this call will return ENOMEDIUM from cg_unified_cached() during the build-time testing, for example when building the package in a Launchpad PPA.

(cherry picked from commit 352ab9d74049b4ac694fdba1a6e67339f12ded93)

Resolves: #2175622


dtardon (approved):
LGTM

PR #380, opened 2023-04-13
(#2156751) Keep udp sockets until we receive a reply or timeout

dtardon:
Resolves: [#2156751](https://bugzilla.redhat.com/show_bug.cgi?id=2156751)


msekletar (approved):
LGTM

PR #384, opened 2023-04-26
(#2186177) pstore: fix crash and forward dummy arguments instead of NULL

msekletar:
[msekleta: in our version of systemd "const char path*" argument of path_join() can't be NULL. Here we don't really want any subdirs paths passed into move_file(), but we can't just pass NULL pointers because they will be forwarded to path_join(). Hence, let's just pass "/" instead.]

RHEL-only

Related: #2186177

### DIFF
--- a/src/pstore/pstore.c
+++ b/src/pstore/pstore.c
@@ -366,7 +366,7 @@ static int run(int argc, char *argv[]) {
 
         /* Move left over files out of pstore */
         for (size_t n = 0; n < list.n_entries; n++)
-                (void) move_file(&list.entries[n], NULL, NULL);
+                (void) move_file(&list.entries[n], "/", "/");
 
         return 0;
 }
--- a/.github/advanced-commit-linter.yml
+++ b/.github/advanced-commit-linter.yml
@@ -1,0 +1,23 @@
+policy:
+  cherry-pick:
+    upstream:
+      - github: systemd/systemd
+      - github: systemd/systemd-stable
+    exception:
+      note:
+        - rhel-only
+  tracker:
+    - keyword:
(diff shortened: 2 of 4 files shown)


src/pstore/pstore.c
```diff
@@ -366,7 +366,7 @@ static int run(int argc, char *argv[]) {
 
         /* Move left over files out of pstore */
         for (size_t n = 0; n < list.n_entries; n++)
-                (void) move_file(&list.entries[n], NULL, NULL);
+                (void) move_file(&list.entries[n], "/", "/");
```
dtardon, line 369: Nice catch! The 3rd arg. could remain `NULL`, but it doesn't matter...

dtardon (dismissed):
LGTM

dtardon:
We need this for 8.7, 8.8 and 8.9 too. Are you going to handle it?

jamacku:
@dtardon I think for 8.7 is too late. 8.8 and 8.9 should be enough.

jamacku (commented):
@msekletar, @dtardon Could we also include patches for the new commit linter:

- https://github.com/redhat-plumbers/systemd-rhel8/commit/a81ace51e1efa2313fcad18f2ef5c6c6cd167240
- https://github.com/redhat-plumbers/systemd-rhel8/commit/7e0eb456a99229d19b8e42491c1ddb7b8050903f

msekletar:
@jamacku I will submit PRs for other branches as well.

jamacku (approved):

msekletar:
> @jamacku I will submit PRs for other branches as well.

@jamacku I think we have PRs for all necessary branches now. PTAL.

PR #385, opened 2023-04-27
(#2190153) pstore: fix crash and forward dummy arguments instead of NULL

msekletar:
Resolves: #2190153

### DIFF
--- a/.github/advanced-commit-linter.yml
+++ b/.github/advanced-commit-linter.yml
@@ -1,0 +1,23 @@
+policy:
+  cherry-pick:
+    upstream:
+      - github: systemd/systemd
+      - github: systemd/systemd-stable
+    exception:
+      note:
+        - rhel-only
+  tracker:
+    - keyword:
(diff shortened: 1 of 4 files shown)


dtardon:
@jamacku Apparently bot has eaten the bug reference in the description.

.github/advanced-commit-linter.yml
```diff
@@ -0,0 +1,23 @@
+policy:
+  cherry-pick:
+    upstream:
+      - github: systemd/systemd
+      - github: systemd/systemd-stable
+    exception:
+      note:
+        - rhel-only
```
jamacku, line 8:
```suggestion
        - rhel-only
        - RHEL-only
```

jamacku:
> @jamacku Apparently bot has eaten the bug reference in the description.

Sorry, `^` and `$` are a part of regex and should not be added to the comment. I'll update GA.

dtardon (approved):
LGTM

PR #386, opened 2023-04-27
(#2190151) pstore: fix crash and forward dummy arguments instead of NULL

msekletar:
[msekleta: in our version of systemd "const char path*" argument of path_join() can't be NULL. Here we don't really want any subdirs paths passed into move_file(), but we can't just pass NULL pointers because they will be forwarded to path_join(). Hence, let's just pass "/" instead.]

RHEL-only

Related: #2190151

### DIFF
--- a/src/pstore/pstore.c
+++ b/src/pstore/pstore.c
@@ -366,7 +366,7 @@ static int run(int argc, char *argv[]) {
 
         /* Move left over files out of pstore */
         for (size_t n = 0; n < list.n_entries; n++)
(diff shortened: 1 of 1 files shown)


dtardon (dismissed):
LGTM

dtardon (approved):
LGTM

mrc0mmand:
~~We still call `move_file()` with `NULL` as the second subdir when going through the ERST dmesg files, so that's going to crash as well, no?~~

Nevermind, I misread the commit message.

PR #387, opened 2023-05-03
(#2190151) test: add a couple of tests for systemd-pstore

mrc0mmand:
Based on:
    - 6858e32d730fd5574eaa3d7fbf4cb12aacaea336
    - edea0d6ac57610b7af603b833b19a846327e3638

Related: #2190151

### DIFF
--- a/src/pstore/pstore.c
+++ b/src/pstore/pstore.c
@@ -181,6 +181,9 @@ static int append_dmesg(PStoreEntry *pe, const char *subdir1, const char *subdir
 
         assert(pe);
 
+        if (arg_storage != PSTORE_STORAGE_EXTERNAL)
+                return 0;
+
         if (pe->content_size == 0)
                 return 0;
 
--- a/src/pstore/pstore.c
+++ b/src/pstore/pstore.c
@@ -207,7 +207,7 @@ static int append_dmesg(PStoreEntry *pe, const char *subdir1, const char *subdir
 static int process_dmesg_files(PStoreList *list) {
         /* Move files, reconstruct dmesg.txt */
         _cleanup_free_ char *erst_subdir = NULL;
-        uint64_t last_record_id = 0;
+        unsigned long long last_record_id = 0;
 
         /* When dmesg is written into pstore, it is done so in small chunks, whatever the exchange buffer
          * size is with the underlying pstore backend (ie. EFI may be ~2KiB), which means an example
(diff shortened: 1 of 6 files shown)


dtardon (approved):
LGTM

jamacku:
> #### The following commits need an inspection
> | commit  | note |
> | --- | --- |
> | [098d1da](https://github.com/redhat-plumbers/systemd-rhel8/commit/098d1da71566163d5bf5ba5524dc3403b2aee8ea) - _test: add a couple of tests for systemd-pstore_  | **Missing upstream reference** :bangbang: |

What about adding a new keyword similar to the `rhel-only` keyword for RHEL-only patches? That would reliably mark commits based on upstream commits that aren't `rhel-only` and don't have to cherry-pick. Benefits would be:
* we would be able to quickly filter those patches
* commit-linter would be able to detect those patches, and it wouldn't block PR.

I was thinking about the `based-on-upstream` keyword. Then commit would look like this:

```
test: add a couple of tests for systemd-pstore

based-on-upstream

Based on:
  - 6858e32d730fd5574eaa3d7fbf4cb12aacaea336
  - edea0d6ac57610b7af603b833b19a846327e3638

Related: #2190151
```

@dtardon, @mrc0mmand, @lnykryn, @msekletar, @jacekmigacz Any thoughts?

jamacku:
Or maybe having a syntax similar to `cherry-pick` and checking the commits against upstream would be better. Something like:

```
test: add a couple of tests for systemd-pstore

(based on commit 6858e32d730fd5574eaa3d7fbf4cb12aacaea336)
(based on commit edea0d6ac57610b7af603b833b19a846327e3638)

Related: #2190151
```

dtardon:
> Or maybe having a syntax similar to `cherry-pick` and checking the commits against upstream would be better. Something like:
> 
> ```
> test: add a couple of tests for systemd-pstore
> 
> (based on commit 6858e32d730fd5574eaa3d7fbf4cb12aacaea336)
> (based on commit edea0d6ac57610b7af603b833b19a846327e3638)
> 
> Related: #2190151
> ```

If we agree to do this, I'd prefer

```
Based-on: 6858e32d730fd5574eaa3d7fbf4cb12aacaea336
Based-on: edea0d6ac57610b7af603b833b19a846327e3638
```

. But I'm not sure we need to complicate this. I'd just mark such "cumulative" commits `RHEL-only` and be done with it. We do this only rarely, after all. I think I've only done it 2-3 times.

Also, what we do more often is to pick just a part of a commit (e.g., just a few hunks from a tree-wide change), but we don't mark these commits in any way (except by adding a comment).

dtardon:
A somewhat related thought: I think we should extend the `RHEL-only` markup in some way. The issue is that right now it means one of two different things:

1. A genuine, intentional divergence from upstream. These changes are likely to be needed in the next RHEL too.
2. A RHEL-specific hack, typically something needed to make CI pass. These changes are typically only relevant for a single major release. We only need to know about them if we need to fix the same issue in a z-stream branch, or if we're doing a rebase (in which case they may or may not be needed, depending on situation).

PR #388, opened 2023-05-04
(#2179309) ci(lint): add new exception keyword for upstream references

jamacku:
for details see: https://github.com/redhat-plumbers/systemd-rhel8/pull/387#issuecomment-1534701242

rhel-only

Related: #2179309

### DIFF
--- a/.github/advanced-commit-linter.yml
+++ b/.github/advanced-commit-linter.yml
@@ -5,7 +5,10 @@ policy:
       - github: systemd/systemd-stable
     exception:
       note:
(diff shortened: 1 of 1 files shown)


msekletar:
@jamacku Is there a reason this is still marked as Draft?

jamacku:
> @jamacku Is there a reason this is still marked as Draft?

I think in the end, we decided with @dtardon  to use just regular `rhel-only` in such cases. So this PR can be closed.

PR #389, opened 2023-05-18
(#2158724) core/mount: escape invalid UTF8 char in dbus reply

msekletar:
When What= or Options= may contain invalid UTF8 chars.

Replaces aaf7b0e41105d7b7cf30912cdac32820f011a219 (#27541).

Resolves: #2158724


dtardon (approved):
LGTM

PR #391, opened 2023-05-24
(#1875056) core/service: when resetting PID also reset known flag

msekletar:
Re-watching pids on cgroup v1 (needed because of unreliability of cgroup empty notifications in containers) is handled bellow at the end of service_sigchld_event() and depends on value main_pid_known flag.

In CentOS Stream 8 container on cgroup v1 the stop action would get stuck indefinitely on unit like this,

$ cat /run/systemd/system/foo.service
[Service]
ExecStart=/bin/bash -c 'trap "nohup sleep 1 & exit 0" TERM;  sleep infinity' ExecStop=/bin/bash -c 'kill -s TERM $MAINPID'
TimeoutSec=0

However, upstream works "fine" because in upstream version of systemd we actually never wait on processes killed in containers and proceed immediately to sending SIGKILL hence re-watching of pids in the cgroup is not necessary. But for the sake of correctness we should merge the patch also upstream.

(cherry picked from commit ff32060f2ed37b68dc26256b05e2e69013b0ecfe)

Resolves: #1875056

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -3249,6 +3249,7 @@ static void service_sigchld_event(Unit *u, pid_t pid, int code, int status) {
                         return;
 
                 s->main_pid = 0;
+                s->main_pid_known = false;
                 exec_status_exit(&s->main_exec_status, &s->exec_context, pid, code, status);
 
                 if (s->main_command) {


msekletar:
Support has sent test packages to the customer. Let's wait with merging until customer confirms that issue is indeed fixed.

msekletar:
Customer confirmed that issue is fixed.

dtardon (approved):
LGTM

mrc0mmand:
This causes a regression in upstream: https://github.com/systemd/systemd/pull/28000.

src/core/service.c
```diff
@@ -3249,6 +3249,7 @@ static void service_sigchld_event(Unit *u, pid_t pid, int code, int status) {
                         return;
 
                 s->main_pid = 0;
+                s->main_pid_known = false;
```
msizanoen1, line 3252: > Re-watching pids on cgroup v1 (needed because of unreliability of cgroup empty notifications in containers) is handled bellow at the end of service_sigchld_event() and depends on value main_pid_known flag.

Can you attempt an alternative solution that doesn't involve marking the main PID as invalid (which breaks `main_pid_good()`)? That doesn't seem correct to me. A better way would be to simply make re-watching PIDs not depend on it.

There are several states `main_pid` can be in:
 - Unset (`!main_pid_known`)
 - Alive (`main_pid_known && main_pid > 0`)
 - Dead (`main_pid_known && main_pid == 0`) (set by `SIGCHLD` handler)

One of the tasks of the `SIGCHLD` handler here is to mark the PID as dead (so `main_pid = 0`). This change breaks that behavior, making systemd treat the main PID as unset instead of dead, which breaks most of the logic that depends on the main PID status, which includes marking a unit as exited or inactive.

msizanoen1 (changes requested):

PR #396, opened 2023-06-07
(#2211358) Fix memory corruption

dtardon:
Resolves: #2211358


msekletar (changes requested):
This backport needs to be reworked. Concretely, introduction of new header file `event-source.h` is fine but it needs to be paired with appropriate removals of type definitions from `sd-event.c`. Now we have everything declared twice which is wrong.

dtardon:
> This backport needs to be reworked. Concretely, introduction of new header file `event-source.h` is fine

Oops, that's a mistake. Apparently I've forgotten to remove it after I'd copied the needed stuff to `sd-event.c`.

msekletar (approved):
LGTM

PR #399, opened 2023-06-29
(RHEL-5991) Do not assign badness to filtered-out syscalls

dtardon:
Resolves: [#2196807](https://bugzilla.redhat.com/show_bug.cgi?id=2196807)

### DIFF
--- a/src/analyze/analyze-security.c
+++ b/src/analyze/analyze-security.c
@@ -480,26 +480,24 @@ static bool syscall_names_in_filter(Set *s, bool whitelist, const SyscallFilterS
         const char *syscall;
 
         NULSTR_FOREACH(syscall, f->value) {
-                bool b;
+                int id;
 
                 if (syscall[0] == '@') {
                         const SyscallFilterSet *g;
+
                         assert_se(g = syscall_filter_set_find(syscall));
-                        b = syscall_names_in_filter(s, whitelist, g);
-                } else {
-#if HAVE_SECCOMP
-                        int id;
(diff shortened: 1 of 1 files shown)


src/analyze/analyze-security.c
```diff
@@ -1538,6 +1538,7 @@ static int assess(const struct security_info *info, Table *overview_table, Analy
                         return log_error_errno(r, "Failed to output table: %m");
         }
 
+        assert(weight_sum > 0);
```
msekletar: We already have assert on the correct place. This commit is not necessary.

src/analyze/analyze-security.c
```diff
@@ -1536,7 +1536,6 @@ static int assess(const struct security_info *info, Table *overview_table, Analy
                         return log_error_errno(r, "Failed to output table: %m");
         }
 
-        assert(weight_sum > 0);
```
msekletar: Because commit which added this line is redundant this is also redundant.

msekletar (changes requested):

src/analyze/analyze-security.c
dtardon: Indeed.

src/analyze/analyze-security.c
dtardon: Blah. Apparently I haven't checked carefully enough.

msekletar (approved):
LGTM

jamacku:
@dtardon could you please rebase this PR, so CentOS CI would run again. Thank you.

PR #401, opened 2023-07-05
(#2158167) Better handle session leader information

msekletar:

### DIFF
--- a/src/login/logind-dbus.c
+++ b/src/login/logind-dbus.c
@@ -3169,6 +3169,9 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
                         session->scope_job = mfree(session->scope_job);
                         (void) session_jobs_reply(session, unit, result);
 
+                        /* Scope job is done so leader should be gone as well. */
+                        session_invalidate_leader(session);
+
                         session_save(session);
                         user_save(session->user);
                 }
--- a/src/login/logind-dbus.c
+++ b/src/login/logind-dbus.c
@@ -784,7 +784,9 @@ static int method_create_session(sd_bus_message *message, void *userdata, sd_bus
                 goto fail;
 
         session_set_user(session, user);
-        session_set_leader(session, leader);
+        r = session_set_leader(session, leader);
+        if (r < 0)
+                goto fail;
 
         session->type = t;
         session->class = c;
--- a/src/login/logind-session.c
+++ b/src/login/logind-session.c
@@ -179,6 +179,23 @@ int session_set_leader(Session *s, pid_t pid) {
         return 1;
 }
 
+int session_invalidate_leader(Session *s) {
+        assert(s);
+
+        if (s->leader <= 0)
+                return 0;
+
+        if (pid_is_alive(s->leader))
+                return 0;
+
+        (void) hashmap_remove_value(s->manager->sessions_by_leader, PID_TO_PTR(s->leader), s);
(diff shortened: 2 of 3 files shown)


msekletar:
@dtardon review of this PR will require extra attention as this is RHEL-only change

src/login/logind-dbus.c
```diff
@@ -3169,6 +3169,8 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
                         session->scope_job = mfree(session->scope_job);
                         (void) session_jobs_reply(session, unit, result);
 
+                        /* Scope job is done so leader must be gone as well. */
```
keszybz: I think this comment is misleading. `session_invalidate_leader()` does not assume that the leader is gone. Should this comment be "Check if the leader is actually gone." ?

src/login/logind-session.c
```diff
@@ -179,6 +179,20 @@ int session_set_leader(Session *s, pid_t pid) {
...
+int session_invalidate_leader(Session *s) {
+        assert(s);
+
+        if (pid_is_alive(s->leader))
+                return 0;
+
+        (void) hashmap_remove_value(s->manager->sessions_by_leader, PID_TO_PTR(s->leader), s);
+        s->leader = 0;
+
+        (void) session_save(s);
```
keszybz, line 194: Maybe this whole block should be guarded by `s->leader > 0`? Do we know that the leader is set if we get to here?

keszybz (commented):
IIUC, this patch will not do anything bad. I haven't looked at this code in a while, so I might be missing obvious things, but I left some comments where I think things can be improved/clarifier.

src/login/logind-dbus.c
msekletar: Maybe the only word that needs changing is "must". Instead it should say "should".

src/login/logind-session.c
msekletar: Good point, we should probably do that.

msekletar:
@keszybz updated. PTAL.

src/login/logind-session.c
```diff
@@ -179,6 +179,23 @@ int session_set_leader(Session *s, pid_t pid) {
         return 1;
 }
 
+int session_invalidate_leader(Session *s) {
+        assert(s);
+
+        if (s->leader <= 0)
```
dtardon, line 185: Maybe use `pid_is_valid()` here?

src/login/logind-session.c
```diff
@@ -1096,6 +1113,7 @@ static int session_dispatch_fifo(sd_event_source *es, int fd, uint32_t revents,
         /* EOF on the FIFO means the session died abnormally. */
 
         session_remove_fifo(s);
+        session_invalidate_leader(s);
```
dtardon, line 1116: Shouldn't this be called from `session_stop()` instead?

src/login/logind-session.c
keszybz: That's probably a bit nicer. Anyway the result is the same.

keszybz (approved):
LGTM.

src/login/logind-session.c
msekletar: It will be called eventually after scope stop job actually finished and then leader process should be gone. Calling it from `session_stop()` seem a bit early to me.

jamacku (approved):
LGTM

PR #404, opened 2023-07-18
(#2223582) test-login: skip consistency checks when logind is not active

jacekmigacz:
There are two ways in swich sd_login_* functions acquire data: some are derived from the cgroup path, but others use the data serialized by logind.

When the tests are executed under Fedora's mock, without systemd-spawn but instead in a traditional chroot, test-login gets confused: the "outside" cgroup path is visible, so sd_pid_get_unit() and sd_pid_get_session() work, but sd_session_is_active() and other functions that need logind data fail.

Such a buildroot setup is fairly bad, but it can be encountered in the wild, so let's just skip the tests in that case.

/* Information printed is from the live system */
sd_pid_get_unit(0, …) → "session-237.scope"
sd_pid_get_user_unit(0, …) → "n/a"
sd_pid_get_slice(0, …) → "user-1000.slice"
sd_pid_get_session(0, …) → "237"
sd_pid_get_owner_uid(0, …) → 1000
sd_pid_get_cgroup(0, …) → "/user.slice/user-1000.slice/session-237.scope" sd_uid_get_display(1000, …) → "(null)"
sd_uid_get_sessions(1000, …) → [0] ""
sd_uid_get_seats(1000, …) → [0] ""
Assertion 'r >= 0' failed at src/libsystemd/sd-login/test-login.c:104, function test_login(). Aborting.

(cherry picked from commit dc400d39b32bf5ad6eefe6ef55f0299cf65787fd)

Resolves: #2223582

### DIFF
--- a/src/libsystemd/sd-login/test-login.c
+++ b/src/libsystemd/sd-login/test-login.c
@@ -115,65 +115,71 @@ static void test_login(void) {
 
         if (session) {
                 r = sd_session_is_active(session);
-                assert_se(r >= 0);
-                log_info("sd_session_is_active(\"%s\") → %s", session, yes_no(r));
+                if (r == -ENXIO)
+                        log_notice("sd_session_is_active() failed with ENXIO, it seems logind is not running.");
+                else {
+                        /* All those tests will fail with ENXIO, so let's skip them. */
 
-                r = sd_session_is_remote(session);
-                assert_se(r >= 0);
-                log_info("sd_session_is_remote(\"%s\") → %s", session, yes_no(r));
-
-                r = sd_session_get_state(session, &state);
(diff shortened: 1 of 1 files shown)


jamacku (changes requested):
@jacekmigacz please update upstream reference (cherry-pick) to: [ac5644635dba54ce5eb0ff394fc0bc772a984849](https://github.com/systemd/systemd/commit/ac5644635dba54ce5eb0ff394fc0bc772a984849)

jamacku (approved):
LGTM

jamacku:
@jacekmigacz, the bug wasn't approved, so it shouldn't be merged.

PR #407, opened 2023-08-04
(#2179309) doc: add downstream CONTRIBUTING document

jamacku:
rhel-only

Related: #2179309

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,67 @@
+# Contributing
+
+Welcome to systemd source-git for CentOS Stream and RHEL. When contributing, please follow the guide below.
+
+## Workflow
+
+```mermaid
+flowchart LR
+    A(Issue) --> B{is fixed\nupstream}
+    B -->|YES| C(backport\nupstream patch)
+    B -->|NO| D(upstream\nsubmit issue or PR)
+    D --> E{accepted\nand fixed}
+    E -->|YES| C
+    E -->|NO| F(rhel-only patch) --> G
+    C --> G(submit PR)
+```
+
+## Filing issues
+
+When you find an issue with systemd used in CentOS Stream or RHEL, please file an issue in [Jira ticket system](https://issues.redhat.com/secure/CreateIssue!default.jspa) (set Project to **RHEL** and Component to **systemd**).
+
+GitHub Issues are not supported tracking system. If your issue is reproducible using the latest upstream version of systemd, please consider creating [upstream issue](https://github.com/systemd/systemd/issues/new/choose).
+
+## Posting Pull Requests
+
+Every Pull Request has to comply with the following rules:
+
+- Each commit has to reference [upstream](https://github.com/systemd/systemd) commit.
+- Each commit has to reference the approved issue/tracker.
+- Pull requests have to pass mandatory CI validation and testing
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+    B -->|NO| D(upstream\nsubmit issue or PR)
+    D --> E{accepted\nand fixed}
+    E -->|YES| C
+    E -->|NO| F(rhel-only patch) --> G
+    C --> G(submit PR)
+```
+
+## Filing issues
+
+When you find an issue with systemd used in CentOS Stream or RHEL, please file an issue by using Bugzilla ticket systemd - [link](https://bugzilla.redhat.com/enter_bug.cgi?classification=Red%20Hat)
```
dtardon: Shouldn't this be Jira instead?

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+
+Resolves: #1234567
+```
+
+### Issue reference
+
+Each commit has to reference the relevant approved systemd issue (see: [Filling issues section](#filing-issues)). For referencing issues, we use the following keywords:
+
+- **Resolves** for commits that directly resolve issues described in a referenced tracker
+- **Relates** for commits related to the referenced issue, but they don't fix it. Usually, tests and documentation.
```
dtardon: s/Relates/Related/

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+- **Reverts** for commits that reverts previously merged commit
```
dtardon, line 52: This section should also explain the format of the reference, not leave it just for an example.

CONTRIBUTING.md
jamacku: You are right. It should be possible to use only Jira for new bugs. That is a good point.

CONTRIBUTING.md
jamacku: Added

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+    B -->|NO| D(upstream\nsubmit issue or PR)
+    D --> E{accepted\nand fixed}
+    E -->|YES| C
+    E -->|NO| F(rhel-only patch) --> G
+    C --> G(submit PR)
+```
+
+## Filing issues
+
+When you find an issue with systemd used in CentOS Stream or RHEL, please file an issue by using Jira ticket system - [link](https://issues.redhat.com/secure/CreateIssue!default.jspa)
```
dtardon: I'd formulate this paragraph as follows:

When you find an issue with systemd used in CentOS Stream or RHEL, please file an issue in [Jira ticket system](https://issues.redhat.com/secure/CreateIssue!default.jspa) (set Project to **RHEL** and Component to **systemd**).

But that's mostly a matter of preference...

dtardon: In any case, the link can be set on the words "Jira ticket system".

jamacku: But I like your preference :) :+1:

jamacku: Done. Thanks for the feedback.

dtardon (approved):
LGTM

PR #417, opened 2023-11-27
(RHEL-1087) test: backport TEST-81-GENERATORS (fstab-generator only)

mrc0mmand:
Some fstab-generator features are not present on RHEL 8 or they behave
differently - in such case there's an inline comment explaining what's
different with a reference to an upstream commit that introduced the
changed behavior.

Realated: https://issues.redhat.com/browse/RHEL-1087

rhel-only

---

Given the recent issue regarding fstab-generator, let's backport the fstab-genreator part of TEST-81-GENERATORS from upstream to make sure we'd catch such issues in the future.


dtardon (approved):
LGTM

PR #418, opened 2023-12-12
(RHEL-5863) service: schedule cleanup of PID hashmaps when we now longer have mai…

msekletar:
…n_pid and we are in container

RHEL-only

Resolves: RHEL-5863

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -34,6 +34,7 @@
 #include "string-table.h"
 #include "string-util.h"
 #include "strv.h"
+#include "virt.h"
 #include "unit-name.h"
 #include "unit.h"
 #include "utf8.h"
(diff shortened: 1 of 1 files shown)


mrc0mmand:
(The C8S fail is unrelated, working on a workaround, will restart the job once it's ready.)

msekletar:
It is not clear at this point whether we end up using this proposed fix or not but I'd appreciate early look from @dtardon. Please let me know what do you think.

msekletar:
Customer confirmed that issue is indeed fix by this cgroup v1 workaround. @dtardon I'd appreciate if you could take a look and give it yay or nay.

dtardon (approved):
LGTM

PR #421, opened 2024-01-12
(RHEL-1087) Automation improvements and issue fixes

jamacku:
* Increase the cron interval to 45 minutes
* Use composite GitHub Action to reduce maintenance cost
* Add Z-streams to an array of supported versions
* Don't fail CI when validation/linting fails; instead show everything in a sticky comment in the PR

#### Example of new status reporting in sticky comment

Screenshot from 2024-01-12 15-38-28


msekletar (approved):
Very superficially LGTM :)

PR #426, opened 2024-02-09
(RHEL-22426) doc: add missing `<listitem>` to `systemd.net-naming-scheme.xml`

jamacku:
rhel-only

Related: RHEL-22426


dtardon (approved):
LGTM

PR #427, opened 2024-02-09
(RHEL-32494) ci: deploy systemd man to GitHub Pages

jamacku:
rhel-only

Related: RHEL-1087

### DIFF
--- a/.github/workflows/deploy-man-pages.yml
+++ b/.github/workflows/deploy-man-pages.yml
@@ -1,0 +1,107 @@
+name: Deploy systemd man to Pages
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - man/*
+      - .github/workflows/deploy-man-pages.yml
+  schedule:
+    # Run every Monday at 4:00 AM UTC
+    - cron: 0 4 * * 1
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  # Single deploy job since we're just deploying
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+
+    permissions:
+      pages: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          RELEASE="$(lsb_release -cs)"
+          sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $RELEASE main restricted universe multiverse' >>/etc/apt/sources.list"
+          sudo add-apt-repository -y ppa:upstream-systemd-ci/systemd-ci
+          sudo apt-get -y update
+          sudo apt-get -y build-dep systemd
+          sudo apt-get install -y \
+            cryptsetup-bin        \
+            expect                \
+            fdisk                 \
+            gettext               \
+            iputils-ping          \
+            isc-dhcp-client       \
+            itstool               \
+            kbd                   \
+            libblkid-dev          \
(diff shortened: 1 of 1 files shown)


dtardon:
@jamacku I guess I'm missing the context here. I understand what the action does, but not why do we want it 🤔

jamacku:
@dtardon, it was requested by @lnykryn. So we would have an easy way to access the `systemd.net-naming-scheme` man page.

I have already done the same for rhel9, so I thought, why not add it also to rhel8. 
https://redhat-plumbers.github.io/systemd-rhel9/

.github/workflows/deploy-man-pages.yml
```diff
@@ -0,0 +1,107 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          RELEASE="$(lsb_release -cs)"
+          sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $RELEASE main restricted universe multiverse' >>/etc/apt/sources.list"
+          sudo add-apt-repository -y ppa:upstream-systemd-ci/systemd-ci
+          sudo apt-get -y update
+          sudo apt-get -y build-dep systemd
+          sudo apt-get install -y \
```
mrc0mmand: This installs a lot of really unnecessary (test-only) deps. The build-dep above should be more than enough for building man pages.

.github/workflows/deploy-man-pages.yml
```diff
@@ -0,0 +1,107 @@
...
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          RELEASE="$(lsb_release -cs)"
+          sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $RELEASE main restricted universe multiverse' >>/etc/apt/sources.list"
+          sudo add-apt-repository -y ppa:upstream-systemd-ci/systemd-ci
```
mrc0mmand: Not necessary.

lnykryn (approved):

.github/workflows/deploy-man-pages.yml
```diff
@@ -0,0 +1,107 @@
...
+      pages: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          RELEASE="$(lsb_release -cs)"
+          sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $RELEASE main restricted universe multiverse' >>/etc/apt/sources.list"
```
mrc0mmand:
```suggestion
          sudo add-apt-repository -y --no-update --enable-source
```

mrc0mmand (changes requested):

.github/workflows/deploy-man-pages.yml
```diff
@@ -0,0 +1,107 @@
...
+            python3-setuptools    \
+            quota                 \
+            strace                \
+            unifont               \
+            util-linux            \
+            zstd                  \
+
+      - name: Build HTML man pages
+        run: |
+          meson build
```
mrc0mmand: Maybe change this to `meson setup build`, newer meson versions complain otherwise.

jamacku:
@mrc0mmand Thanks for review. I needed to make it work fast for Lukaš to share it. I'll try to minimize the list of dependencies.

dtardon:
> @dtardon, it was requested by @lnykryn. So we would have an easy way to access the `systemd.net-naming-scheme` man page.

Such way does already exist: `./build/man/man` and `./build/man/html`. These commands build and display a specified man page, e.g., `./build/man/man systemd.net-naming-scheme`. They have the additional advantage that one can see the effect of local changes immediately.

jamacku:
true, but it was for sharing with others, as a reference for technical writers.

jamacku:
Oops, for some reason, the origin was pointing to upstream :disappointed:

.github/workflows/deploy-man-pages.yml
jamacku: Thank you for review @mrc0mmand

PR #428, opened 2024-02-14
(RHEL-6210) cryptsetup: try to detach the device multiple times before giving up

lnykryn:
Terrible hack for RHEL8. We are missing the blockdev@.target, so we can't correctly express some dependencies. That means when systemd-cryptsetup detach can fail, because for example a filesystem is still mounted on top of the device. This can cause shutdown to hang forever. See https://issues.redhat.com/browse/RHEL-6210 So as a workaround, let's try to detach the device multiple times before giving up. */

RHEL-only
Resolves: RHEL-6210

### DIFF
--- a/src/cryptsetup/cryptsetup.c
+++ b/src/cryptsetup/cryptsetup.c
@@ -772,7 +772,18 @@ int main(int argc, char *argv[]) {
 
                 crypt_set_log_callback(cd, cryptsetup_log_glue, NULL);
 
-                r = crypt_deactivate(cd, argv[2]);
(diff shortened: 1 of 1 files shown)


lnykryn:
Marking as a draft so it is not automatically merge. Maybe we will be able to backport blockdev@.target

lnykryn:
Hmm with encrypted root, there is now 10 seconds delay in shutdown every time :crying_cat_face:

jamacku:
Replaced by #429

Issue #434, opened 2024-03-06
Outstanding CI issue


mrc0mmand:
(I'm opening this mainly as a reminder to myself, so I don't debug the same stuff twice.)

In https://github.com/redhat-plumbers/systemd-rhel8/pull/433 CI started failing with:

```
 120/298 test-fs-util                                       FAIL              0.05s   killed by signal 6 SIGABRT
>>> SYSTEMD_KBD_MODEL_MAP=/build/src/locale/kbd-model-map SYSTEMD_LANGUAGE_FALLBACK_MAP=/build/src/locale/language-fallback-map MALLOC_PERTURB_=236 PATH=/build/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /build/build/test-fs-util
 ✀  
stderr:
Assertion 'touch_file(a, false, test_mtime, test_uid, test_gid, 0640) >= 0' failed at ../src/test/test-fs-util.c:549, function test_touch_file(). Aborting.
```

The fail itself is caused by `touch_file()` calling `chmod()` on a symlink, which then returns `EOPNOTSUPP`. I suspect this is something that has changed in later kernels (and in this case it affects CI, because it runs in a docker container on Ubuntu Jammy). On C8S kernel this seems to work, but only on `tmpfs`:

```c
# cat main.c 
#include <assert.h>
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>

int main(int argc, char *argv[]) {
        int fd, r;
        char fdpath[64];

        if (argc < 2)
                return 1;

        fd = open(argv[1], O_PATH|O_CLOEXEC|O_NOFOLLOW);
        if (fd < 0) {
                fprintf(stderr, "open() failed: %m\n");
                return 1;
        }

        sprintf(fdpath, "/proc/self/fd/%i", fd);
        r = chmod(fdpath, 0640);
        printf("chmod(%s)=%d (%d)\n", fdpath, r, errno);

        return r;
}
```

```
# mount -t tmpfs tmpfs /tmp
# (cd /tmp; ln -s symlink dangling; ln -s /etc/os-release symlink)
# (cd /var/tmp; ln -s symlink dangling; ln -s /etc/os-release symlink)
# gcc -o main main.c -D_GNU_SOURCE
# ./main /tmp/dangling
chmod(/proc/self/fd/3)=0 (0)
[...]

PR #436, opened 2024-03-13
test: skip the symlink part of test_touch_file() in GH Actions

mrc0mmand:
Our (RHEL 8) touch_file() is not clever enough and does chmod() on a symlink, which fails with EOPNOTSUPP on newer kernels. This is not an issue on the RHEL 8 kernel, where doing chmod() on a symlink works (albeit only on tmpfs) but in GH Actions we run in a container, and with the underlying kernel doing chmod() on a symlink fails even on tmpfs:

RHEL 8:
```
~# mount -t tmpfs tmpfs /tmp
~# (cd /tmp; ln -s symlink dangling; ln -s /etc/os-release symlink) ~# (cd /var/tmp; ln -s symlink dangling; ln -s /etc/os-release symlink) ~# gcc -o main main.c -D_GNU_SOURCE
~# ./main /tmp/dangling
chmod(/proc/self/fd/3)=0 (0)
~# ./main /tmp/symlink
chmod(/proc/self/fd/3)=0 (0)
~# ./main /var/tmp/dangling
chmod(/proc/self/fd/3)=-1 (95)
~# ./main /var/tmp/symlink
chmod(/proc/self/fd/3)=-1 (95)
```

Newer kernel:
```
~# uname -r
6.7.4-200.fc39.x86_64
~# ./main /tmp/dangling
chmod(/proc/self/fd/3)=-1 (95)
~# ./main /tmp/symlink
chmod(/proc/self/fd/3)=-1 (95)
~# ./main /var/tmp/dangling
chmod(/proc/self/fd/3)=-1 (95)
~# ./main /var/tmp/symlink
chmod(/proc/self/fd/3)=-1 (95)
```

Backporting the necessary patches would be way too risky so late in the RHEL 8 cycle, so let's just skip the offending test when running in a container. This should work around this issue in GH Actions, but still run the test on C8S/RHEL8 machines.

See: #434

RHEL-only

---

Opening this without any tracker, since there might not be a next minor release, but we still might need the "fix" for z-streams.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -18,6 +18,9 @@ jobs:
     steps:
       - name: Repository checkout
         uses: actions/checkout@v1
+      # FIXME: drop once https://github.com/actions/runner-images/issues/9491  is resolved
+      - name: Reduce ASLR entropy
+        run: sudo sysctl -w vm.mmap_rnd_bits=28
       - name: Install build dependencies
         run: sudo -E .github/workflows/unit_tests.sh SETUP
       - name: Build & test (${{ env.CENTOS_RELEASE }} / ${{ matrix.phase }})
--- a/src/test/test-fs-util.c
+++ b/src/test/test-fs-util.c
@@ -15,6 +15,7 @@
 #include "stdio-util.h"
 #include "string-util.h"
 #include "strv.h"
+#include "tests.h"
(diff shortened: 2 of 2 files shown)


mrc0mmand:
The first commit should be needed only for the next couple of days, until the GH Actions runners are fixed.

msekletar (approved):

PR #438, opened 2024-04-23
(RHEL-30645) resolved: never return ::1 when localhost or local hostname is resolved while IPv6 is off in the kernel

jacekmigacz:


jamacku:
The issue has been closed, so let's close this PR as well.

PR #440, opened 2024-06-26
(RHEL-1087) ci: point C8S containers to the Vault

mrc0mmand:
Temporarily point repos in C8S containers to the Vault (since C8S is EOL), until we figure out a _proper_ solution.

Related: RHEL-1087


jamacku (approved):
LGTM

PR #441, opened 2024-11-05
(RHEL-19215) logind: don't setup idle session watch for lock-screen and greeter

msekletar:
Reason to skip the idle session logic for these session classes is that
they are idle by default.

(cherry picked from commit 508b4786e8592e82eb4832549f74aaa54335d14c)

Resolves: RHEL-19215


mrc0mmand:
CI fails are known issues (#434, #440).

jamacku (approved):
LGTM

PR #444, opened 2025-01-28
(RHEL-76308) core: fix member access within null pointer

jamacku:
config_parse_tasks_max() is also used for parsing system.conf or user.conf. In that case, userdata is NULL.

Fixes #10362.

(cherry picked from commit 958b8c7bd7a3cf1e710faf8c19a528cc94c214fe)

Resolves: RHEL-76308


mrc0mmand (approved):

mrc0mmand:
The CI fails are known issues (#434 and oversaturated CentOS CI infra).

PR #446, opened 2025-03-05
(RHEL-81056) man: explicitly document that "reboot -f" is different from "systemctl reboot -f"

jamacku:


jamacku:
Failure is unrelated to this change.

lnykryn (approved):

PR #449, opened 2025-09-02
test-execute: let's ignore the difference between CLD_KILLED and CLD_DUMPED

mrc0mmand:
Depending on system configuration and whether SCMP_ACT_KILL_PROCESS or SCMP_ACT_KILL_THREAD is available/used processes might coredump on specific coredumps or are just plain killed. For our test case the difference doesn't really matter, hence let's hide it away.

(cherry picked from commit c3ab2c389ee60d92fb8d7fe779ae9c4e3c092e4c)

### DIFF
--- a/src/test/test-execute.c
+++ b/src/test/test-execute.c
@@ -30,6 +30,12 @@
 
 typedef void (*test_function_t)(Manager *m);
 
+static int cld_dumped_to_killed(int code) {
+        /* Depending on the system, seccomp version, … some signals might result in dumping, others in plain
+         * killing. Let's ignore the difference here, and map both cases to CLD_KILLED */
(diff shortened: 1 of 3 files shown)


mrc0mmand:
This PR backports changes to fix the issue mentioned in https://github.com/redhat-plumbers/systemd-rhel8/pull/295#issuecomment-3236599754. I also did some changes on CentOS CI's side to work around #434, so now the CIs should be, hopefully green.

This fix is not important by itself (it's just a test-only change), so let's just pick it up in the next z-stream batch, if any.

mrc0mmand:
Merged as part of https://github.com/redhat-plumbers/systemd-rhel8/pull/450.

PR #450, opened 2025-09-29
(RHEL-108744) timer: don't run service immediately after restart of a timer

mrc0mmand:
When a timer is restarted, don't reset the last_trigger field.
This prevents the timer from triggering immediately.

Fixes: #31231
(cherry picked from commit 3fc44a0f68412b649e16f12ff2f97a36c615457d)

Resolves: RHEL-108744

---

Note: this also pulls in changes from #449 to fix a CI fail.

### DIFF
--- a/src/core/timer.c
+++ b/src/core/timer.c
@@ -601,8 +601,6 @@ static int timer_start(Unit *u) {
         if (r < 0)
                 return r;
 
-        t->last_trigger = DUAL_TIMESTAMP_NULL;
-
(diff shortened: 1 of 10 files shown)


mrc0mmand:
The fail in TEST-19-DELEGATE is unrelated (caused by #295; fix is still being evaluated). The follow-up is just a cosmetic change, not important here.

jamacku (approved):
LGTM

PR #451, opened 2025-10-02
(RHEL-9322) Revert test for a change we don't have in RHEL-8

dtardon:
Revert "test: extend testcase to ensure controller membership doesn't regress"

This reverts commit a3ce5bf1ff531f29c329b3298f3ef69d99e3cbed.

Related: RHEL-9322


mrc0mmand:
The commit message is missing `rhel-only`.

The failing `Unit tests` job is getting fixed in #450, so just ignore that fail for now.

jamacku:
> The commit message is missing `rhel-only`.

It's ok for Reverts of upstream commits.

jamacku (approved):
LGTM

PR #452, opened 2025-10-02
(RHEL-118835) run: update checks to allow running with a user's bus

jelly:
systemd-run is documented to as being able to connect and run on a specific user bus with "--user --machine=lennart@.host" arguments. This PR updates some logic that prevented this from working.

(cherry picked from commit cbdc29492097e24ef3320280bc2a8dedbce02d9a)

Resolves: RHEL-118835


msekletar (approved):
LGTM

PR #453, opened 2025-10-31
(RHEL-38859) cryptsetup-generator: continue parsing after error

mrc0mmand:
Let's make the crypttab parser more robust and continue even if parsing
of a line failed.

(cherry picked from commit 83813bae7ae471862ff84b038b5e4eaefae41c98)

Resolves: RHEL-38859

---

I skipped the other two commits from https://github.com/systemd/systemd/pull/33100 as they don't seem to be needed to resolve RHEL-38859, but I can pull them in if needed.


dtardon (approved):
LGTM

PR #457, opened 2025-12-04
(RHEL-118835) Revert "(RHEL-118835) run: update checks to allow running with a user's bus"

jamacku:
Reverts redhat-plumbers/systemd-rhel8#452


mrc0mmand (approved):
LGTM

PR #458, opened 2025-12-05
(RHEL-132317) logind: fix crash in logind on user-specified message string

mrc0mmand:
This is trivially exploitable (in the sense of causing a crash from SEGV) e.g. by 'shutdown now "Message %s %s %n"'. The message is settable through polkit, but is limited to auth_admin:
```
<action id="org.freedesktop.login1.set-wall-message">
         <description gettext-domain="systemd">Set a wall message</description>
         <message gettext-domain="systemd">Authentication is required to set a wall message</message>
         <defaults>
                <allow_any>auth_admin_keep</allow_any>
                <allow_inactive>auth_admin_keep</allow_inactive>
                <allow_active>auth_admin_keep</allow_active>
        </defaults>
</action>
```
Bug introduced in 9ef15026c0e7e6600372056c43442c99ec53746e ('logind/systemctl: introduce SetWallMessage and --message', 2015-09-15).

(cherry picked from commit 0cb09bcb825ab86ba4ca70be4e6322eaf9baee95)

Related: RHEL-132317

PR #46, opened 2019-12-02
(#1723722) debug-generator: enable custom systemd.debug_shell tty

jsynacek:
(cherry picked from commit 93912e872fb14e9c372e090409e429084a6450f5)

Resolves: #1723722

PR #463, opened 2026-01-16
(RHEL-133353) encountering ordering cycle triggers segmentation fault

tcornell-bus:
Backport RHEL-8.10 RHEL-132317 first [PR](https://github.com/redhat-plumbers/systemd-rhel8/pull/456):
The generated string may include %, which will confuse both the xprintf call, and the VA_FORMAT_ADVANCE macro.

Pass the generated string as an argument to a "%s" format string instead.

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel8/commit/7325a2b2d15af09a9389723d6153050130c0bd36)

Resolves: RHEL-133353

---
Backport RHEL-8.10 RHEL-132317 second [PR](https://github.com/redhat-plumbers/systemd-rhel8/pull/458):
This is trivially exploitable (in the sense of causing a crash from SEGV) e.g. by 'shutdown now "Message %s %s %n"'. The message is settable through polkit, but is limited to auth_admin:
```
<action id="org.freedesktop.login1.set-wall-message">
         <description gettext-domain="systemd">Set a wall message</description>
         <message gettext-domain="systemd">Authentication is required to set a wall message</message>
         <defaults>
                <allow_any>auth_admin_keep</allow_any>
                <allow_inactive>auth_admin_keep</allow_inactive>
                <allow_active>auth_admin_keep</allow_active>
        </defaults>
</action>
```
Bug introduced in https://github.com/redhat-plumbers/systemd-rhel8/commit/9ef15026c0e7e6600372056c43442c99ec53746e ('logind/systemctl: introduce SetWallMessage and --message', 2015-09-15).

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel8/commit/0cb09bcb825ab86ba4ca70be4e6322eaf9baee95)

Related: RHEL-133353

### DIFF
--- a/src/core/transaction.c
+++ b/src/core/transaction.c
@@ -341,6 +341,9 @@ static char* merge_unit_ids(const char* unit_log_field, char **pairs) {
                 size += next + 1;
         }
 
+        if (!ans)
+                return strdup("");
+
         return TAKE_PTR(ans);
 }
 
--- a/src/core/transaction.c
+++ b/src/core/transaction.c
@@ -395,7 +395,7 @@ static int transaction_verify_order_one(Transaction *tr, Job *j, Job *from, unsi
                                    j->unit->id,
                                    unit_id == array ? "ordering cycle" : "dependency",
                                    *unit_id, *job_type,
-                                   unit_ids);
+                                   "%s", unit_ids);
 
                 if (delete) {
                         const char *status;
@@ -404,7 +404,7 @@ static int transaction_verify_order_one(Transaction *tr, Job *j, Job *from, unsi
                                    "MESSAGE=%s: Job %s/%s deleted to break ordering cycle starting with %s/%s",
                                    j->unit->id, delete->unit->id, job_type_to_string(delete->type),
                                    j->unit->id, job_type_to_string(j->type),
-                                   unit_ids);
+                                   "%s", unit_ids);
 
                         if (log_get_show_color())
                                 status = ANSI_HIGHLIGHT_RED " SKIP " ANSI_NORMAL;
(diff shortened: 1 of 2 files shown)


tcornell-bus:
I see the commits to move jobs from Travis CI to GH Actions are not in this branch. I don't mind trying to apply them here, but I wanted to check before doing so. What do you think @mrc0mmand?

mrc0mmand:
> I see the commits to move jobs from Travis CI to GH Actions are not in this branch.

Whoah, what a blast from the past...

> I don't mind trying to apply them here, but I wanted to check before doing so. What do you think @mrc0mmand?

Having at least some working CI here would be nice. But I wouldn't bother backporting all the necessary commits one by one in this case as there's probably _a lot_ of them - just copy-over the `.github/workflows/unit_tests.{sh,yml}` files from the `rhel-8.4.0` branch and commit them as a single "rhel-only" commit (we've done this a couple of times already across multiple RHEL branches in similar scenarios). Hopefully that'd be all that's needed.

mrc0mmand:
(Just a note that the config you're copying is using Ubuntu 20.04 which won't work, I guess you're copying it from a 8.4.0 branch that doesn't yet contain changes from https://github.com/redhat-plumbers/systemd-rhel8/pull/46. So I'd copy the workflow-related files _with_ changes from that PR included).

tcornell-bus:
:facepalm: hopefully I did it right this time!

mrc0mmand:
```
Run sudo -E .github/workflows/unit_tests.sh SETUP
sudo: .github/workflows/unit_tests.sh: command not found
Error: Process completed with exit code 1
```

I have a sneaking suspicion that the `.github/workflows/unit_tests.sh` file is not executable :)

tcornell-bus:
Do I need this commit https://github.com/redhat-plumbers/systemd-rhel8/commit/ed4bd948f28a1f811f617b1e8dfb741b28d5be83? 
```
man/meson.build:181:0: ERROR: Program 'rsync' not found
```

mrc0mmand:
Looks like this might be more painful than I originally anticipated. I'd probably drop the CI stuff for now, and ship only the fixes - we run both the unit & integration tests internally anyway and, more importantly, on a "correct" RHEL version, so they should, hopefully, work there.

I can look into fishing out the necessary CI patches later (and without pressure), so when the time for 8.2.z comes again (hopefully it won't), we would have at least one working CI here.

mrc0mmand (approved):
LGTM, thanks!

jamacku:
Released under tag https://github.com/redhat-plumbers/systemd-rhel8/releases/tag/v239-31.11

PR #464, opened 2026-01-21
(RHEL-138710) core: only activate transaction that contain useful jobs

lnykryn:
If no real jobs were added to the transaction, do not activate it. The JOB_NOP anchor does not perform any useful work and activating such transaction only wastes resources.

Fixes #9751


jamacku (changes requested):
@lnykryn, please fix the upstream reference by using `git cherry-pick -x`.

lnykryn:
@jamacku oops sorry, it seems I have an old version of my cherry-pick script on my other machine

Should be fixed now

jamacku (approved):

PR #465, opened 2026-03-11
(RHEL-152085) core: validate input cgroup path more prudently

dtardon:

Resolves: RHEL-152085


msekletar (approved):
path normalization check prevents us from reaching problematic alloca(). I guess that is good enough for RHEL-8.

PR #468, opened 2026-04-16
(RHEL-163867) core/manager: fix memory leak

msekletar:
Follow-up for 63899267ba57ab4b0b11c039f5c6b03a78528aa7

RHEL-only

Resolves: RHEL-163867


dtardon (approved):
LGTM

PR #470, opened 2026-05-04
(RHEL-5986) execute: make sure to call into PAM after initializing resource limits

msekletar:
We want that pam_limits takes precedence over our settings, after all.

Fixes: #11386
(cherry picked from commit ce932d2d331886190dca9d54571d3d99d9fe8cfc)

Resolves: RHEL-5986


dtardon (approved):
LGTM

PR #286, opened 2022-05-31
(#2040247) Return ENOSYS for unknown system calls

dtardon:
Resolves: [#2040247](https://bugzilla.redhat.com/show_bug.cgi?id=2040247)

PR #473, opened 2026-05-19
(RHEL-159158) core: validate input cgroup path more prudently

tcornell-bus:

Resolves: RHEL-159158

### DIFF
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -456,6 +456,12 @@ static int method_get_unit_by_control_group(sd_bus_message *message, void *userd
         if (r < 0)
                 return r;
 
+        if (!path_is_absolute(cgroup))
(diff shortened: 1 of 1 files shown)


mrc0mmand (approved):
LGTM, thanks!

As for the CI fail - CentOS CI failed in all tests run via nspawn because in RHEL 8.6 and older we're missing https://github.com/redhat-plumbers/systemd-rhel8/pull/286. However, backporting this is both non-trivial and risky, so I tweaked the CentOS CI configuration to run the tests only under QEMU/KVM.

With that the only failing test now is parse-hwdb, because we're also missing https://github.com/redhat-plumbers/systemd-rhel8/pull/454/ - that should be possible to backport relatively easily to 8.6, but that can be done later in the next batch, if there's any.

jamacku:
released under tag https://github.com/redhat-plumbers/systemd-rhel8/releases/tag/v239-58.16

PR #461, opened 2026-01-15
(RHEL-133354) encountering ordering cycle triggers segmentation fault

tcornell-bus:
Backport RHEL-8.10 RHEL-132317 first [PR](https://github.com/redhat-plumbers/systemd-rhel8/pull/456):
The generated string may include %, which will confuse both the xprintf call, and the VA_FORMAT_ADVANCE macro.

Pass the generated string as an argument to a "%s" format string instead.

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel8/commit/7325a2b2d15af09a9389723d6153050130c0bd36)

Resolves: RHEL-133354

---
Backport RHEL-8.10 RHEL-132317 second [PR](https://github.com/redhat-plumbers/systemd-rhel8/pull/458):
This is trivially exploitable (in the sense of causing a crash from SEGV) e.g. by 'shutdown now "Message %s %s %n"'. The message is settable through polkit, but is limited to auth_admin:
```
<action id="org.freedesktop.login1.set-wall-message">
         <description gettext-domain="systemd">Set a wall message</description>
         <message gettext-domain="systemd">Authentication is required to set a wall message</message>
         <defaults>
                <allow_any>auth_admin_keep</allow_any>
                <allow_inactive>auth_admin_keep</allow_inactive>
                <allow_active>auth_admin_keep</allow_active>
        </defaults>
</action>
```
Bug introduced in https://github.com/redhat-plumbers/systemd-rhel8/commit/9ef15026c0e7e6600372056c43442c99ec53746e ('logind/systemctl: introduce SetWallMessage and --message', 2015-09-15).

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel8/commit/0cb09bcb825ab86ba4ca70be4e6322eaf9baee95)

Related: RHEL-133354

PR #474, opened 2026-05-19
(RHEL-159157) core: validate input cgroup path more prudently

tcornell-bus:

Resolves: RHEL-159157

### DIFF
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -456,6 +456,12 @@ static int method_get_unit_by_control_group(sd_bus_message *message, void *userd
         if (r < 0)
                 return r;
 
+        if (!path_is_absolute(cgroup))
+                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Control group path is not absolute: %s", cgroup);
+
+        if (!path_is_normalized(cgroup))
+                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Control group path is not normalized: %s", cgroup);
+
         u = manager_get_unit_by_cgroup(m, cgroup);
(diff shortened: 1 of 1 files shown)


tcornell-bus:
To try and save you some time, the failure looks to be the same as last rhel-8.4 PR with this [waiver](https://github.com/redhat-plumbers/systemd-rhel8/pull/461#pullrequestreview-3671539444):

> The test-fs-util fail is a known issue (https://github.com/redhat-plumbers/systemd-rhel8/issues/434), and CentOS CI doesn't apply here.

mrc0mmand:
> To try and save you some time, the failure looks to be the same as last rhel-8.4 PR with this [waiver](https://github.com/redhat-plumbers/systemd-rhel8/pull/461#pullrequestreview-3671539444):
> 
> > The test-fs-util fail is a known issue (#434), and CentOS CI doesn't apply here.

Yup, indeed. I was trying to fix (or workaround) CentOS CI to make it work even here and it seems to finally work. Apart from the same nspawn/seccomp commits I mentioned in https://github.com/redhat-plumbers/systemd-rhel8/pull/473#pullrequestreview-4328612905 we're also missing ca86de228e19cea268ec3eeabc9097d7c28fbf24 but we can't push it into this branch, as libmicrohttpd on RHEL 8.4 is older than what the commit needs. So I worked around it on CentOS CI side which cherry-picks the commit only for the CI runs.

And we don't need any fixes for parse-hwdb here, as it seems to be in good-enough shape in this branch.

mrc0mmand (approved):
LGTM, thanks!

jamacku:
released under tag https://github.com/redhat-plumbers/systemd-rhel8/releases/tag/v239-45.17

PR #477, opened 2026-07-22
(RHEL-112550) core: allow manager_serialize() to fail correctly

siteshwar:
If manager_serialize() fails in the middle (which it hopefully doesn't) make sure to fix up m->n_reloading correctly again so that we don't leave it > 0 when it really shouldn't be.

(cherry picked from commit 4daf832afaeac471085e56445a05218217bf5107)

Resolves: RHEL-112550


dtardon (approved):
LGTM

PR #479, opened 2026-08-13
(RHEL-219224) mount: regenerate all deps whenever a mount's parameters  changes

siteshwar:
Whenever we notice a change on an existing /proc/self/mountinfo line,
let's update the deps generated from it. For that, let's flush out the
old deps generated this way, and add in the new ones.

This takes benefit of the fact that today (unlike a comment this patch
removes says) we can remove deps in a somewhat reasonable way.

(cherry picked from commit a37422045fbb68ad68f734e5dc00e0a5b1759773)

Resolves: RHEL-219224

### DIFF
--- a/src/core/mount.c
+++ b/src/core/mount.c
@@ -1425,6 +1425,32 @@ static int mount_dispatch_timer(sd_event_source *source, usec_t usec, void *user
         return 0;
 }
 
+static int update_parameters_proc_self_mount_info(
+                Mount *m,
+                const char *what,
+                const char *options,
+                const char *fstype) {
+
+        MountParameters *p;
+        int r, q, w;
+
+        p = &m->parameters_proc_self_mountinfo;
+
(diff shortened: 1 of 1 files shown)


siteshwar:
I have not been able to verify these changes work as expected locally, but based on my discussion with @mrc0mmand I am marking this request as ready for review.

msekletar (changes requested):
NACK, actually merging this would cause regression that was later addressed upstream in https://github.com/systemd/systemd/issues/11342.

msekletar:
@siteshwar please avoid proposing changes to core parts of the systemd (specially for RHEL-8 which is in maintenance phase 2) if you don't have full mental model of what these changes will cause otherwise such haste changes may have catastrophic unforeseen consequences.

siteshwar:
Thanks for the review!

PR #480, opened 2026-08-17
(RHEL-26842) pid1: set SYSTEMD_NSS_DYNAMIC_BYPASS=1 env var for dbus-daemon

siteshwar:
There's currently a deadlock between PID 1 and dbus-daemon: in some cases dbus-daemon will do NSS lookups (which are blocking) at the same time PID 1 synchronously blocks on some call to dbus-daemon. Let's break that by setting SYSTEMD_NSS_DYNAMIC_BYPASS=1 env var for dbus-daemon, which will disable synchronously blocking varlink calls from nss-systemd to PID 1.

In the long run we should fix this differently: remove all synchronous calls to dbus-daemon from PID 1. This is not trivial however: so far we had the rule that synchronous calls from PID 1 to the dbus broker are OK as long as they only go to interfaces implemented by the broke itself rather than services reachable through it. Given that the relationship between PID 1 and dbus is kinda special anyway, this was considered acceptable for the sake of simplicity, since we quite often need metadata about bus peers from the broker, and the asynchronous logic would substantially complicate even the simplest method handlers.

This mostly reworks the existing code that sets SYSTEMD_NSS_BYPASS_BUS= (which is a similar hack to deal with deadlocks between nss-systemd and dbus-daemon itself) to set SYSTEMD_NSS_DYNAMIC_BYPASS=1 instead. No code was checking SYSTEMD_NSS_BYPASS_BUS= anymore anyway, and it used to solve a similar problem, hence it's an obvious piece of code to rework like this.

Issue originally tracked down by Lukas Märdian. This patch is inspired and closely based on his patch:

       https://github.com/systemd/systemd/pull/22038

Fixes: #15316
Co-authored-by: Lukas Märdian <slyon@ubuntu.com>
(cherry picked from commit de90700f36f2126528f7ce92df0b5b5d5e277558)

Resolves: RHEL-26842

### DIFF
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -1688,11 +1688,11 @@ static int build_environment(
                 our_env[n_env++] = x;
         }
 
-        /* If this is D-Bus, tell the nss-systemd module, since it relies on being able to use D-Bus look up dynamic
-         * users via PID 1, possibly dead-locking the dbus daemon. This way it will not use D-Bus to resolve names, but
-         * check the database directly. */
-        if (p->flags & EXEC_NSS_BYPASS_BUS) {
-                x = strdup("SYSTEMD_NSS_BYPASS_BUS=1");
+        /* If this is D-Bus, tell the nss-systemd module, since it relies on being able to use blocking
+         * Varlink calls back to us for look up dynamic users in PID 1. Break the deadlock between D-Bus and
+         * PID 1 by disabling use of PID1' NSS interface for looking up dynamic users. */
+        if (p->flags & EXEC_NSS_DYNAMIC_BYPASS) {
+                x = strdup("SYSTEMD_NSS_DYNAMIC_BYPASS=1");
                 if (!x)
                         return -ENOMEM;
                 our_env[n_env++] = x;
--- a/src/core/execute.h
+++ b/src/core/execute.h
@@ -287,7 +287,7 @@ typedef enum ExecFlags {
         EXEC_NEW_KEYRING       = 1 << 3,
         EXEC_PASS_LOG_UNIT     = 1 << 4, /* Whether to pass the unit name to the service's journal stream connection */
         EXEC_CHOWN_DIRECTORIES = 1 << 5, /* chown() the runtime/state/cache/log directories to the user we run as, under all conditions */
-        EXEC_NSS_BYPASS_BUS    = 1 << 6, /* Set the SYSTEMD_NSS_BYPASS_BUS environment variable, to disable nss-systemd for dbus */
+        EXEC_NSS_DYNAMIC_BYPASS    = 1 << 6, /* Set the SYSTEMD_NSS_DYNAMIC_BYPASS environment variable, to disable nss-systemd blocking on PID 1, for use by dbus-daemon */
         EXEC_CGROUP_DELEGATE   = 1 << 7,
         EXEC_IS_CONTROL        = 1 << 8,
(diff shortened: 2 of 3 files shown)


siteshwar:
Manually verified that the fix works on RHEL 8.

msekletar (changes requested):
I suspect this is likely wrong thing to do because nss-systemd in RHEL-8 doesn't use Varlink at all so there is no point in avoiding blocking Varlink calls. As for avoiding D-Bus calls from dbus-daemon context this is already covered by EXEC_NSS_BYPASS_BUS.

In other words, patch is NOOP for RHEL-8 and I don't think you have correct reproducer.

siteshwar:
I used  the reproducer from [RHEL-26842](https://redhat.atlassian.net/browse/RHEL-26842?focusedCommentId=17969166):
```
# Configuration:
# - samba-winbind-4.24.5-100.el10 installed
# - nsswitch.conf: passwd: files winbind systemd
# - winbind running in standalone mode
# - dbus-broker-36-4.el10 (default)
#
# Result: dbus restart completed in 0.099s with no deadlock
# No errors in journal, all services healthy after restart

$ time systemctl restart dbus
real  0m0.099s

$ systemctl status dbus --no-pager | head -5
  dbus-broker.service - D-Bus System Message Bus
     Loaded: loaded
     Active: active (running)

$ getent passwd root
root:x:0:0:Super User:/root:/bin/bash
$ getent passwd nobody
nobody:x:65534:65534:Kernel Overflow User:/:/usr/sbin/nologin
```

msekletar:
@siteshwar Well, that is AI generated reproducer so you should take it with a grain of salt...

msekletar:
@siteshwar Also, if you look at the comment in the reproducer it points to packages with el10 dist-tag and you are proposing change for RHEL-8

siteshwar:
> @siteshwar Well, that is AI generated reproducer so you should take it with a grain of salt...

It is also mentioned in [this comment](https://redhat.atlassian.net/browse/RHEL-26842?focusedCommentId=5013058) from the reporter and the `SYSTEMD_NSS_DYNAMIC_BYPASS`variable is [documented](https://github.com/redhat-plumbers/systemd-rhel8/blob/d489e6c6315d20f695d6ed269de15fb3c204d082/doc/ENVIRONMENT.md?plain=1#L113) for RHEL 8.

In other words, I do not know if this the correct reproducer for the original issue. But I was able reproduce the hang on RHEL 8 and the changes in this pull request appeared to fix it (as I tried multiple times to reproduce, but could not reproduce).

msekletar:
> In other words, I do not know if this the correct reproducer for the original issue. But I was able reproduce the hang on RHEL 8 and the changes in this pull request appeared to fix it (as I tried multiple times to reproduce, but could not reproduce).

But I think that is just coincidence. If you look at nss-systemd in RHEL-8 you will see that if SYSTEMD_NSS_DYNAMIC_BYPASS is set then nss-systemd short-circuits and doesn't do any user lookup, but again on RHEL-8 there is NO Varlink based lookup that would be blocking, instead nss-systemd is looking into auxiliary files in /run directly and that shouldn't be blocking.

Anyway this should be properly debugged first before we jump to conclusions. If you have a reliable reproducer for RHEL-8 this should be fairly easy to debug.

siteshwar:
@msekletar Thanks for the review!

